PR #1, opened 2018-12-09
Add simple example

z3ntu:
The license is still from the example from the legacy branch but as it was basically completely rewritten it can be anything^^

Also I noticed that nearly all files in the project don't have any copyright header - which should probably be added.


chin123:
Is this going to be merged? It was pretty helpful for me, and it would be great if it was in the main repo.

z3ntu:
I've updated the example for the newest mautrix-go release.

PR #10, opened 2020-06-22
Add crypto SQL store from mautrix-whatsapp

nikofil:
* Adds a `SQLCryptoStore` class for storing cryptographic material in a database.

Signed-off-by: Nikos Filippakis <me@nfil.dev>


nikofil:
I'm not sure about the `CreateTables` method, I was thinking that maybe I could add a migrations system like the one you have in `mautrix-whatsapp` only for the crypto tables but if they won't change it probably won't be necessary. WDYT?

tulir:
I usually handle all database migrations at the "top level" rather than inside libraries, but I guess other people might not like that. 

Maybe add a `crypto_version` table and insert `1` there in `CreateTables`? That way it doesn't need the actual migration system code yet, but it should be easy to add later if needed.

nikofil:
I started doing that, then realised the `GetVersion` function would also need to create the version table to work before the first migration could run so there's not much point in creating the version table in the migration itself. So I ended up doing nothing, besides restructuring the `CreateTables` function a bit. :D We can add that code if we need to add migrations.

I'd also like to add some comments to the `OlmMachine` functions to make it more clear how they can be used, if you don't mind.

crypto/sql_store.go
```diff
@@ -0,0 +1,507 @@
...
+	"github.com/lib/pq"
+	"github.com/pkg/errors"
+
+	"maunium.net/go/mautrix/crypto/olm"
+	"maunium.net/go/mautrix/id"
+)
+
+// WarnlnLogger is used to log warnings.
+type WarnlnLogger interface {
+	Warnln(...interface{})
```
tulir, line 23: This should probably support the same logger as OlmMachine

tulir:
Added some comments to all the crypto stuff

crypto/sql_store.go
nikofil: Sure, done.

PR #12, opened 2020-06-24
Crypto store and olm machine tests

nikofil:
* Add tests for the functionality of the Gob and SQL crypto stores
* Change `RemoveOutboundGroupSession` in the Gob store to not check if a session is shared, to be consistent with the SQL store
* Add a test for the functionality of `OlmMachine`, creating an olm and megolm session, sharing the megolm session and encrypting a message

### DIFF
--- a/crypto/machine_test.go
+++ b/crypto/machine_test.go
@@ -1,0 +1,140 @@
+// Copyright (c) 2020 Nikos Filippakis
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package crypto
+
+import (
+	"os"
+	"testing"
+
+	"maunium.net/go/mautrix"
+	"maunium.net/go/mautrix/event"
+	"maunium.net/go/mautrix/id"
+)
+
+type emptyLogger struct{}
+
+func (emptyLogger) Error(message string, args ...interface{}) {}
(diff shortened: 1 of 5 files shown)


nikofil:
Hopefully these are useful, if you'd like tests for something else let me know.
I plan to look at some of the TODOs in the code next, particularly in the crypto part, if that doesn't conflict with what you're doing.

crypto/machine_test.go
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) 2020 Tulir Asokan
```
tulir, line 1: You should put your name in things you wrote

nikofil: Sure, changed

crypto/store_test.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		t.Fatalf("Error opening db: %v", err)
+	}
+	sqlStore := NewSQLCryptoStore(db, "sqlite3", id.DeviceID("dev"), []byte("test"), emptyLogger{})
+	if err = sqlStore.CreateTables(); err != nil {
+		t.Fatalf("Error creating tables: %v", err)
+	}
+
+	os.Remove("gob_store_test.gob")
+	gobStore, err := NewGobStore("gob_store_test.gob")
```
tulir, line 37: This doesn't seem to get removed after the tests

nikofil: Right, the last test was leaving its gob. Fixed that by also returning a cleanup func from `getCryptoStores`.

PR #15, opened 2020-07-16
Forward room key request & response

nikofil:
* Add ability to ask other devices to forward a room key
* Handle forwarded room keys by saving them to the crypto store when coming from the same user ID
* Handle requests for forwarding room keys from other devices
* Add some types that will be needed for key verification

Currently keys are shared with any device that belongs to the same user ID, however that should change once verification is working to only allow verified devices.

### DIFF
--- a/crypto/machine.go
+++ b/crypto/machine.go
@@ -7,9 +7,6 @@
 package crypto
 
 import (
-	"context"
-	"math/rand"
-	"strconv"
 	"sync"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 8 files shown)


crypto/machine.go
```diff
@@ -180,15 +190,192 @@ func (mach *OlmMachine) HandleToDeviceEvent(evt *event.Event) {
...
+					// close channel to notify listener that the key was received
+					close(ch.(chan struct{}))
+				}
+			}
+			// TODO handle m.dummy encrypted to-device event
 		}
-		// TODO handle other unencrypted to-device events. At least m.room_key_request and m.verification.start
+	case *event.RoomKeyRequestEventContent:
+		// Only forward keys to the same user ID
+		if content.Action == event.KeyRequestActionRequest && mach.Client.UserID == evt.Sender {
```
tulir, line 204: This should probably check device verification status somewhere. It can have a flag to disable the check (like `AllowUnverifiedDevices` does for encrypting messages) at least until device verification is actually implemented

crypto/machine.go
```diff
@@ -180,15 +190,192 @@ func (mach *OlmMachine) HandleToDeviceEvent(evt *event.Event) {
...
+		return err
+	}
+
+	// get Olm session
+	olmSess, err := mach.CryptoStore.GetLatestSession(device.IdentityKey)
+	if err != nil {
+		return err
+	}
+	if olmSess == nil {
+		return err
```
tulir, line 287: `err` is nil at this point, so it should probably return a different error. (is this case even possible in practice?)

crypto/machine.go
```diff
@@ -180,15 +190,192 @@ func (mach *OlmMachine) HandleToDeviceEvent(evt *event.Event) {
...
+		},
+	}
+	// send messages to the devices
+	if _, err := mach.Client.SendToDevice(event.ToDeviceRoomKeyRequest, toDeviceReq); err != nil {
+		return nil, err
+	}
+	resChan := make(chan bool, 1)
+	go func() {
+		// when exiting, send a message to all devices cancelling this key request
+		defer func() {
```
tulir, line 344: Is there any reason for this to be a `defer` rather than just at the end of the function? It doesn't look like anything should panic and there aren't any early returns.

crypto/machine.go
```diff
@@ -180,15 +190,192 @@ func (mach *OlmMachine) HandleToDeviceEvent(evt *event.Event) {
...
+			},
+		},
+	)
+
+	return err
+}
+
+// RequestRoomKey sends a key request for a room to the current user's devices. If the context is cancelled, then so is the key request.
+// Returns a bool channel that will get notified either when the key is received or the request is cancelled.
+func (mach *OlmMachine) RequestRoomKey(ctx context.Context, toUser id.UserID, toDevice id.DeviceID,
```
tulir, line 308: It would probably be a good idea to split `RequestRoomKey`, `importForwardedRoomKey` and maybe the `*event.RoomKeyRequestEventContent` handling to a separate `keysharing.go` file. I'm not completely sure where `SendEncryptedToDevice` should be

crypto/machine.go
nikofil: It really shouldn't be. Maybe I should just remove the check and let it panic further on if this happens? It would mean something went very wrong.

nikofil: I think this entire `if` was wrong to be there, actually. Removed.

crypto/machine.go
nikofil: A defer made more sense in my original design of this function tbh (when I thought that we might receive cancellation messages from many devices, before reading that if they don't want to share they just ignore the message). I guess I just like defers.

Moved after the `select`.

crypto/machine.go
nikofil: Good idea, moved. I kept `SendEncryptedToDevice` there however as it's supposed to be a generic utility function.

crypto/machine.go
nikofil: Done (hardcoded to true for now).

PR #16, opened 2020-07-24
SAS verification

nikofil:
* SAS verification with both decimal and emoji methods (but not legacy)
* Configurable via OlmMachine attributes that decide whether a verification should be started and if the two SAS strings match


crypto/verification_sas_methods.go
```diff
@@ -0,0 +1,143 @@
...
+	acceptUserID id.UserID, acceptDeviceID id.DeviceID, transactionID string, sas *olm.SAS) ([7]VerificationEmoji, error) {
+
+	sasInfo := "MATRIX_KEY_VERIFICATION_SAS" +
+		initUserID.String() + initDeviceID.String() +
+		acceptUserID.String() + acceptDeviceID.String() +
+		transactionID
+
+	var emojis [7]VerificationEmoji
+	sasBytes, err := sas.GenerateBytes([]byte(sasInfo), 6)
+	emojisStr := ""
```
tulir, line 123: Is `emojisStr` used for anything else than debugging? If not, you could probably just log the `emojis` array instead of building a string manually

crypto/verification.go
```diff
@@ -0,0 +1,589 @@
...
+
+	mach.Log.Error("emoji %v", useEmojiSAS)
+
+	go mach.timeoutAfter(device.UserID, transactionID, timeout)
+
+	return nil
+}
+
+// StartSASVerification sends the SAS verification start message to another device.
+func (mach *OlmMachine) StartSASVerification(toUserID id.UserID, toDeviceID id.DeviceID, transactionID string, useEmoji bool) (*event.VerificationStartEventContent, error) {
```
tulir, line 499: Does this need to be exported? i.e. is there any case where someone would want to use this over `NewSASVerificationWith`?

crypto/verification.go
```diff
@@ -0,0 +1,589 @@
...
+		}
+	} else {
+		mach.Log.Debug("Not accepting SAS verification %v from %v of user %v", content.TransactionID, otherDevice.DeviceID, otherDevice.UserID)
+		mach.CancelSASVerification(otherDevice.UserID, otherDevice.DeviceID, content.TransactionID, "Not accepted by user", event.VerificationCancelByUser)
+	}
+}
+
+// NewSASVerificationWith starts the SAS verification process with another device.
+// If the transaction ID is empty, a new one is generated.
+func (mach *OlmMachine) NewSASVerificationWith(device *DeviceIdentity, transactionID string, timeout time.Duration, useEmojiSAS bool) error {
```
tulir, line 455: It would probably be nicer if the three last arguments were in an optional struct (see how [`SendMessageEvent`](https://github.com/tulir/mautrix-go/blob/master/client.go#L587) has an "optional" `extra` parameter). Also, timeout should have some default value and emojis should be on by default (maybe it should be disableEmojis instead of useEmojis)

crypto/machine.go
```diff
@@ -36,9 +37,18 @@ type OlmMachine struct {
 
 	AllowUnverifiedDevices bool
 
+	// AcceptVerificationFrom determines whether the machine will accept verification requests from this device.
+	AcceptVerificationFrom func(*DeviceIdentity) bool
+	// VerifySASNumbersMatch should return whether the given SAS numbers match with the numbers displayed on other device.
+	VerifySASNumbersMatch func([3]uint, *DeviceIdentity) bool
+	// VerifySASNumbersMatch should return whether the given SAS emojis match with the emojis displayed on other device.
+	// This is only needed if support for emoji SAS verification is wanted.
+	VerifySASEmojisMatch func([7]VerificationEmoji, *DeviceIdentity) bool
```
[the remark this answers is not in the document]
tulir: Also, in addition to those hooks, there should be other hooks for verification cancelled, verification finished, etc

crypto/machine.go
```diff
@@ -36,9 +37,18 @@ type OlmMachine struct {
 
 	AllowUnverifiedDevices bool
 
+	// AcceptVerificationFrom determines whether the machine will accept verification requests from this device.
+	AcceptVerificationFrom func(*DeviceIdentity) bool
```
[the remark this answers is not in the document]
nikofil: If this returns false, the transaction will be cancelled. Do you mean it could ie. return a second value that says whether it should be cancelled or simply ignored? When would we want to just ignore it?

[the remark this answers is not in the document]
tulir: At least if you're ignoring a user, you wouldn't want to respond to verification requests at all. Not sure if to-device events are filtered from /sync when you're ignoring someone server-side, but an option to ignore requests locally would probably be good anyway

crypto/verification.go
```diff
@@ -0,0 +1,596 @@
...
+
+	verState.startEventCanonical = string(canonical)
+	_, loaded := mach.keyVerificationTransactionState.LoadOrStore(device.UserID.String()+":"+transactionID, verState)
+	if loaded {
+		return errors.New("Transaction already exists")
+	}
+
+	mach.Log.Error("emoji %v", useEmojiSAS)
+
+	go mach.timeoutAfter(device.UserID, transactionID, timeout)
```
tulir, line 500: The timeouts for each step should probably be separated, e.g. if you wait for X-1 seconds for the other person to accept the request, you should have more than 1 second to compare the emoji.

crypto/verification_sas_methods.go
```diff
@@ -0,0 +1,143 @@
...
+// VerificationEmoji describes an emoji that might be sent for verifying devices.
+type VerificationEmoji struct {
+	Emoji       rune
+	Description string
+}
+
+// GetSASVerificationNumbers generates the three numbers that need to match with the other device for a verification to be valid.
+func (mach *OlmMachine) GetSASVerificationNumbers(initUserID id.UserID, initDeviceID id.DeviceID,
+	acceptUserID id.UserID, acceptDeviceID id.DeviceID, transactionID string, sas *olm.SAS) ([3]uint, error) {
+	sasInfo := "MATRIX_KEY_VERIFICATION_SAS" +
```
tulir, line 92: The `curve25519` key agreement method is apparently deprecated, the new one (`curve25519-hkdf-sha256`) is different: matrix-org/matrix-doc#2630 / https://matrix.org/docs/spec/client_server/unstable#hkdf-calculation

tulir: We probably don't need to support the old method at all:

> New implementations are discouraged from implementing the `curve25519` method.

crypto/verification.go
tulir: `CancelSASVerification` seems to be similar (might not need to be exported)

nikofil: I was thinking that these could be used if someone wants to provide their own implementation for handling SAS to save them some effort, as the current one is coupled to the default syncer implementation (the handle* methods are unexported). The exported methods don't make any reference to the `verificationState` struct either, which is used by the handle* methods.

crypto/machine.go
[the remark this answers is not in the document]
nikofil: Resolving as I see you did this already.

crypto/verification_sas_methods.go
nikofil: Indeed, removed

crypto/verification.go
nikofil: Renamed it to `SendSASVerificationStart` / `SendSASVerificationCancel` etc and made the comment a bit more explicit.

nikofil:
I cleaned up the `useEmoji` mess, hooks now include a method that returns the list of SAS methods in order of preference and the verification uses the first one that is shared with the other device.

crypto/verification.go
tulir: Looks like you renamed `StartSASVerification` to `SendSASVerificationStart`, which is better. `CancelSASVerification` should probably be renamed to `SendSASVerificationCancel`, so that the manual methods all start with `SendSASVerification`.

Related to canceling, a separate method for canceling a verification started with the `New` method might be useful.

nikofil: I thought I also renamed `CancelSASVerification`, huh. Done now.

Do you think doing user cancellation here using a context is a good idea? I'm thinking we could also `WithTimeout` on top of the given context instead of `time.Sleep(timeout) and then cancel transaction`. Then we could also have a `WithCancel` so whenever the transaction progresses, we cancel the current timeout context and start a new one with the same timeout. Or am I complicating it too much?

nikofil: I didn't like how the context thing was turning out, so I added a new `CancelSASVerification` instead for cancelling the transaction by the user.

crypto/verification.go
nikofil: Done, whenever the transaction progresses it should now start a new timeout.

PR #18, opened 2020-08-21
Cross signing and SSSS

nikofil:
I thought I'd create this PR ahead of time so that I don't make a huge one again out of nowhere.

Implemented:
 * Basic SSSS functionalities
 * Fetch x-signing key seeds from user's SSSS
 * Store other users' uploaded keys and signatures to be able to check them later
 * libolm bindings for PkSigning objects
 * Use the seeds from SSSS to restore the x-signing keys
 * Create and upload x-signing keys to SSSS if needed
 * Verify a user / device is trusted by following the trust chain
 * Create and upload signatures for other keys
 * Sign other user's keys after SAS and have them sign ours
 * Delete old signatures when keys are rotated
 * In-room verification


crypto/utils/utils.go
```diff
@@ -0,0 +1,133 @@
...
+	return
+}
+
+// GenA256CTRIV generates a random IV for AES256-CTR with the last bit set to zero.
+func GenA256CTRIV() (iv [AESCTRIVLength]byte) {
+	_, err := rand.Read(iv[:])
+	if err != nil {
+		panic(err)
+	}
+	iv[8] &= 0x7F
```
tulir, line 64: Shouldn't this modify index 7 (8th byte) and use `0x7E` instead of `0x7F` to set bit 63 to 0?

```suggestion
	iv[7] &= 0b11111110
```

(The master branch has the same code for making the IV for key exports)

crypto/utils/utils.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// GenAttachmentA256CTR generates a new random AES256-CTR key and IV suitable for encrypting attachments.
+func GenAttachmentA256CTR() (key [AESCTRKeyLength]byte, iv [AESCTRIVLength]byte) {
+	_, err := rand.Read(key[:])
+	if err != nil {
+		panic(err)
+	}
+
+	// For some reason we leave the 8 last bytes empty even though AES256-CTR has a 16-byte block size.
```
tulir, line 50:
```suggestion
	// The last 8 bytes of the IV act as the counter in AES-CTR, which means they're left empty here
```

crypto/utils/utils.go
nikofil: I wasn't sure myself, I wrote this judging by https://github.com/matrix-org/matrix-js-sdk/blob/048d7a9b6377c155c56d5981e9b5d32e060398d2/src/crypto/aes.js#L50

tulir: I'm not sure either, but I thought `iv[8] &= 0x7F` sets bit 64 (65th bit), while `iv[7] &= 0xFE` sets bit 63 (64th bit). Maybe we should ask someone who actually knows how bits and byte arrays work

tulir: To add to the confusion, Element Android [seems to do](https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/secrets/DefaultSharedSecretStorageService.kt#L269-L272) `iv[9] &= 0x7F`?!

tulir: Apparently bits are weird and Android/iOS are doing it wrong, but the js-sdk way is correct

crypto/utils/utils.go
nikofil: Changed in another commit

nikofil:
This is not WIP anymore. You can see https://github.com/nikofil/go-neb/commit/d4410bd0d37e8d8d7dea832b4ebba85774266f60 for an example of how to try out the new cross-signing and in-room verification.

Essentially you need to:
* call `olmMachine.RetrieveCrossSigningKeysWithPassphrase("passphrase")` once
* call `olmMachine.ProcessInRoomVerification(event, relatesTo)` when receiving a message event with msgtype=`m.key.verification.request` or any other in-room `m.key.verification.*` message (if the message is encrypted, use the relates-to of the encrypted event)

tulir:
> if the event is encrypted, use the relates-to of the encrypted event as the parameter and not the decrypted one

[DecryptMegolmEvent](https://github.com/tulir/mautrix-go/blob/master/crypto/decryptmegolm.go#L83-L86) should already handle copying it (and EncryptMegolmEvent copies it in the other direction)

nikofil:
Right, I implemented that interface now instead, looks like it still works.

Note by the way that I remove a couple lines when JSONifying relates-to (https://github.com/tulir/mautrix-go/pull/18/files#diff-0698c4af6d3b534b889ebba23c8ed3ee) as that was messing with generating the commitment - not sure what might break because of that.

Spazzlo:
Is this close to getting merged?

tulir:
The verification stuff doesn't work, but I guess I could merge it anyway

PR #20, opened 2020-11-09
Update the Registration struct

babolivier:
Add the `protocols` option. Also add `omitempty` to the `rate_limited` option since it's not required to appear in the registration file, and so we fall back to the homeserver's default value (for example Synapse's is `true` so if one wants to keep to default behaviours they probably don't want it to be automatically set to `false`).

### DIFF
--- a/appservice/registration.go
+++ b/appservice/registration.go
@@ -21,7 +21,7 @@ type Registration struct {
 	AppToken        string     `yaml:"as_token"`
 	ServerToken     string     `yaml:"hs_token"`
 	SenderLocalpart string     `yaml:"sender_localpart"`
-	RateLimited     bool       `yaml:"rate_limited,omitempty"`
+	RateLimited     *bool       `yaml:"rate_limited,omitempty"`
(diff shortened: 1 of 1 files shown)


appservice/registration.go
```diff
@@ -21,9 +21,10 @@ type Registration struct {
 	AppToken        string     `yaml:"as_token"`
 	ServerToken     string     `yaml:"hs_token"`
 	SenderLocalpart string     `yaml:"sender_localpart"`
-	RateLimited     bool       `yaml:"rate_limited"`
+	RateLimited     bool       `yaml:"rate_limited,omitempty"`
```
tulir, line 24: Won't this omit `false`? It probably needs to be a pointer to allow not setting the value

babolivier: Oh yes it would, good point.

babolivier: Fixed in https://github.com/tulir/mautrix-go/pull/20/commits/ae55ddbaec64041b0d663c9d9ff3348603e0d066

tulir (approved):

PR #21, opened 2020-11-27
Annotate RelatesTo struct with json field names

rkfg:
Without these reactions don't work

### DIFF
--- a/event/relations.go
+++ b/event/relations.go
@@ -22,9 +22,9 @@ const (
 )
 
 type RelatesTo struct {
-	Type    RelationType
-	EventID id.EventID
-	Key     string
+	Type    RelationType `json:"rel_type"`
+	EventID id.EventID   `json:"event_id"`
+	Key     string       `json:"key"`
 }
 
 type serializableInReplyTo struct {


tulir:
The annotations are on the lines below, so that replies can be handled the same way as other relations

rkfg:
I don't understand. Without these annotations the library sends wrong events that are not parsed by Element so the reaction is invisible. It looks like this: `{"m.relates_to":{"Type":"m.annotation","EventID":"$iqMXve2BvUmksWTD9F8hQvpHQsfJJ_QVdc0-n45ldTQ","Key":"🔒️"}}` while it should be `{"m.relates_to":{"rel_type":"m.annotation","event_id":"$iqMXve2BvUmksWTD9F8hQvpHQsfJJ_QVdc0-n45ldTQ","key":"🔒️"}}`. I also found out that Synapse doesn't accept `Key` with the uppercase 'k' (this has nothing to do with your library, just information):
```
Traceback (most recent call last):
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/rest/client/v1/room.py", line 259, in on_POST
    requester, event_dict, txn_id=txn_id
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/handlers/message.py", line 736, in create_and_send_nonmember_event
    requester, event_dict, txn_id=txn_id
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/handlers/message.py", line 528, in create_event
    auth_event_ids=auth_event_ids,
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/util/metrics.py", line 92, in measured_func
    r = await func(self, *args, **kwargs)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/handlers/message.py", line 838, in create_new_client_event
    aggregation_key = relation["key"]
KeyError: 'key'
```
So this has to be fixed.

tulir:
Reactions work fine in gomuks. How are you sending the reaction exactly?

rkfg:
The code is like this (b is my struct, b.client is `*mautrix.Client`):
```go
	msg, err := b.client.SendText(roomID, "Some announcement")
	if err != nil {
		log.Println(err)
		return
	}
	_, err = b.client.SendReaction(roomID, msg.EventID, "🔒️")
```
The message appears, the reaction does not. With the proposed patch all works fine. Do they work for you in Element? Can you check with a logger that HTTP requests are correct? I did this to see them:
```go
	logger := maulogger.DefaultLogger
	logger.PrintLevel = 0
	b.client.Logger = logger
```
I really don't understand what you meant by "The annotations are on the lines below", shouldn't they be in the structure itself to be properly serialized to JSON? Because by default it uses the field names that don't conform to the spec imo.

rkfg:
I see, you must be referring to `serializableRelatesTo`. But [it's not used in SendReaction](https://github.com/tulir/mautrix-go/blob/dcc56f7fc25815c878ffe12cba7fcbd1fa12348f/client.go#L698). Maybe it should be replaced with `serializableRelatesTo` there then?

tulir:
617e6c94cc3a2f046434bf262fadd993daf02141 might fix it

rkfg:
Yes, it works now, thanks! The fix is not obvious at all.

tulir:
The `SendMessageEvent` doc does say

> contentJSON should be a pointer to something that can be encoded as JSON using json.Marshal

I just hadn't remembered to update the other methods to follow the isntructions :D

PR #26, opened 2021-02-20
Add client wrapper functions for Device Management endpoints

edwargix:
https://matrix.org/docs/spec/client_server/r0.6.1#id73

Signed-off-by: David Florness <david@florness.com>


edwargix:
I put everything after the `sendToDevice` definitions because as of r0.6.1, "Send-to-Device Messaging" is section 3.9 and "Device Management" is section 3.10.

tulir (approved):
Looks good, thanks!

PR #36, opened 2021-08-25
Add health endpoint

Half-Shot:
This adds a simple /health endpoint that responds with "OK".


tulir:
Would `/_matrix/mau/live` be fine to match what [mautrix-python has](https://github.com/mautrix/python/blob/master/mautrix/appservice/appservice.py#L101-L102)? https://github.com/mautrix/go/commit/386ac2e0c62ea8271f03121b25cad90c0d5b5ff2

Half-Shot:
Sure, works for me :)

PR #44, opened 2021-11-12
Retry on 429s (too many requests) from Matrix server

ownaginatious:
I noticed that when backfilling messages after my `mautrix-imessage` has been offline for a while, many of the messages get lost due to 429s from my Matrix server.

This PR attempts to resolve that by adding 429 to the list of status codes that are retried by the client. The optional `Retry-After` header that may come back with it is parsed and overrides the default backoff time if present to ideally prevent subsequent 429s. If no `Retry-After` header is present, then the default `backoff` is used.

Question: if a 429 is returned, should the remaining retry count still be reduced? I'd argue no and just keep retrying until the 429 goes away as that insinuates that any future requests made are more likely to also fail anyway.

### DIFF
--- a/client.go
+++ b/client.go
@@ -517,6 +517,27 @@ func (cli *Client) handleResponseError(req *http.Request, res *http.Response) ([
 	}
 }
 
+// parseBackoffFromResponse extracts the backoff time specified in the Retry-After header if present. See
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After.
+func (cli *Client) parseBackoffFromResponse(res *http.Response, now time.Time, fallback time.Duration) time.Duration {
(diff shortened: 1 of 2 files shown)


tulir:
In general you should disable ratelimiting, but I guess it might be useful to support auto-retrying on 429 anyway

ownaginatious:
> In general you should disable ratelimiting, ...

I guess so, but given that it is enabled in the default `synapse` configuration, I'm guessing it's something that clients/bridges should expect and attempt to honor? 🤷‍♂️

tulir:
It's disabled by default for bridges, so bridges won't expect it. The only exception is double puppeting.

ownaginatious:
Ah, so that's why it seems like only my messages get dropped when backfilling. Thanks!

tulir (approved):

PR #45, opened 2021-11-15
Add support for message send checkpoints

sumnerevans:


appservice/message_send_checkpoint.go
```diff
@@ -0,0 +1,102 @@
...
+	RemoteStep     MessageSendCheckpointStep = "REMOTE"
+	CommandStep    MessageSendCheckpointStep = "COMMAND"
+)
+
+type MessageSendCheckpointStatus string
+
+const (
+	SuccesssStatus    MessageSendCheckpointStatus = "SUCCESS"
+	WillRetryStatus   MessageSendCheckpointStatus = "WILL_RETRY"
+	PermFailureStatus MessageSendCheckpointStatus = "PERM_FAILURE"
```
tulir, line 37: `Step` and `Status` should be before the name of the step/status in the constant names

appservice/message_send_checkpoint.go
```diff
@@ -0,0 +1,102 @@
...
+	defer cancel()
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, endpoint, &body)
+	if err != nil {
+		return err
+	}
+
+	req.Header.Set("Authorization", "Bearer "+asToken)
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
```
tulir, line 89: We should probably add the proper user agent header for these requests (from `mautrix.DefaultUserAgent`, and maybe some extra string like "checkpoint sender" like https://github.com/mautrix/go/blob/master/client.go#L107)

tulir (approved):

PR #54, opened 2022-01-01
Add `/context` API

qua3k:


client.go
```diff
@@ -1296,6 +1296,28 @@ func (cli *Client) Messages(roomID id.RoomID, from, to string, dir rune, limit i
...
+// See https://spec.matrix.org/v1.1/client-server-api/#get_matrixclientv3roomsroomidcontexteventid
+func (cli *Client) Context(roomID id.RoomID, eventID id.EventID, limit int) (resp *RespContext, err error) {
+	filter := cli.Syncer.GetFilterJSON(cli.UserID)
+	filterJSON, err := json.Marshal(filter)
+	if err != nil {
+		return nil, err
+	}
+	query := map[string]string{
+		"filter": string(filterJSON),
+	}
```
tulir, line 1311: The filter would probably be better as a parameter, it doesn't have to be the same as the sync filter (and it's also not required)

tulir (approved):

PR #57, opened 2022-01-03
Add sticker sending support

qua3k:


tulir:
You should probably just send it manually, the media sending wrappers are old and not particularly useful (no media info, no encryption support).

PR #58, opened 2022-01-03
util: add ffmpeg utils for converting bytes and files

sumnerevans:


util/ffmpeg.go
```diff
@@ -0,0 +1,136 @@
...
+//
+// Returns: the converted data
+func ConvertBytes(data []byte, outputExtension string, inputArgs []string, outputArgs []string, inputMime string) ([]byte, error) {
+	tempdir, err := ioutil.TempDir("", "mautrix_ffmpeg_*")
+	if err != nil {
+		return nil, err
+	}
+	defer os.RemoveAll(tempdir)
+
+	tempfile, err := ioutil.TempFile(tempdir, "*"+ExtensionFromMimetype(inputMime))
```
tulir, line 115: Why does this use a tempfile inside the existing tempdir? Just `input.extension` should be fine

util/ffmpeg.go
```diff
@@ -0,0 +1,136 @@
...
+//
+// Args:
+// * data: The media data to convert
+// * outputExtension: The extension that the output file should be.
+// * inputArgs: Arguments to tell ffmpeg how to parse the input file.
+// * outputArgs: Arguments to tell ffmpeg how to convert the file to reach the wanted output.
+// * inputMime: The mimetype of the input data.
+//
+// Returns: the converted data
+func ConvertBytes(data []byte, outputExtension string, inputArgs []string, outputArgs []string, inputMime string) ([]byte, error) {
```
tulir, line 108: These should probably either be moved to `util/ffmpeg/ffmpeg.go` so they're used as `ffmpeg.ConvertBytes`, or renamed to `util.ConvertBytesFFmpeg`

(although `ExtensionFromMimetype` doesn't really belong in `util/ffmpeg/` if the convert methods are moved there)

util/ffmpeg.go
```diff
@@ -0,0 +1,136 @@
...
+
+	log.Info(args)
+
+	cmd := exec.Command("ffmpeg", args...)
+	vcLog := log.Sub("ffmpeg").Writer(log.LevelWarn)
+	cmd.Stdout = vcLog
+	cmd.Stderr = vcLog
+	err := cmd.Run()
+	if err != nil {
+		return "", fmt.Errorf("ffmpeg error: %+v. stdout: %s. stderr: %s", err, cmd.Stdout, cmd.Stderr)
```
tulir, line 88: I don't think this works, it just returns an error message containing pointers to the log writer. `cmd.StdoutPipe()` would work for getting the stream data as errors, but it might be easier to just return the code and look at the logs if something goes wrong

```suggestion
		return "", fmt.Errorf("ffmpeg error: %w", err)
```

tulir (changes requested):

util/ffmpeg.go
sumnerevans: The advantage is that we can just nuke the directory after we are done which deletes both the input and output files.

tulir: I mean `{tempdir}/input.{extension}` should be fine as the input path like mautrix-python does, there's no reason to make a `ioutil.TempFile` inside the `tempdir`

sumnerevans: Oh, I see what you mean. Will change.

tulir (approved):

PR #59, opened 2022-01-06
Add initial ACL support

qua3k:


event/type.go
```diff
@@ -175,6 +175,7 @@ var (
 	StateTopic             = Type{"m.room.topic", StateEventType}
 	StateRoomAvatar        = Type{"m.room.avatar", StateEventType}
 	StatePinnedEvents      = Type{"m.room.pinned_events", StateEventType}
+	StateServerAcls        = Type{"m.room.server_acl", StateEventType}
```
tulir, line 178: The event type is singular so this should be too, and [initialisms should have a consistent case in Go](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

```suggestion
	StateServerACL        = Type{"m.room.server_acl", StateEventType}
```

(also needs to be changed where the event type is used and in the struct name)

tulir (approved):

PR #64, opened 2022-02-01
make olm build on FreeBSD

kcbark:


tulir:
I think the correct way is setting the `LIBRARY_PATH` environment variable rather than hardcoding paths if olm is installed in a weird location

PR #65, opened 2022-02-16
Fix mismatching arguments in scan

S7evinK:


tulir:
I should probably find a query builder or linter :thinking:

I think I'll change the query to match how `GetDevice` works

PR #73, opened 2022-04-01
Add GetAliases call to Client

hifi:
This could also be called `GetRoomAliases` but this fit with the rest of the alias functions.


responses.go
```diff
@@ -293,6 +293,9 @@ type RespAliasResolve struct {
 	RoomID  id.RoomID `json:"room_id"`
 	Servers []string  `json:"servers"`
 }
+type RespAliasList struct {
```
sumnerevans, line 296:
```suggestion

type RespAliasList struct {
```

tulir (approved):

PR #74, opened 2022-04-06
Use EscapedPath instead of RawPath

abaker:


abaker:
url.String() calls url.EscapedPath(), and in there it compares the result of `unescape(u.RawPath, encodePath)` to u.Path. But u.RawPath was missing the domain so the unescape result did not match. Then it would try escaping u.Path using encodePath mode, which doesn't escape / and ;

PR #79, opened 2022-06-14
More issues caught by linter

qua3k:


qua3k:
Could setup golangci-lint as a more effective linter in the future

PR #88, opened 2022-09-02
MSC3870: allow media create endpoint to specify alternate upload URL

thefinn93:
[MSC3870](https://github.com/Fizzadar/matrix-spec-proposals/blob/media-async-upload-url/proposals/3870-media-async-upload-url.md) extends the async media uploads spec to allow the server to specify a URL to send the media to.

There is another related change to indicate to the server that we will follow redirects when downloading media


tulir (approved):
Looks fine

PR #90, opened 2022-09-14
event/powerlevels: make {users,events}_default not omitempty

sumnerevans:


tulir:
Those fields default to 0 if unspecified, so specifying 0 shouldn't be necessary?

sumnerevans:
This is for making sure that it's encoded properly when sending to clients. These fields are required.

sumnerevans:
:scream: https://github.com/matrix-org/complement/blob/main/tests/csapi/apidoc_room_state_test.go#L77

I guess I'll make a PR over there to make it just check that it's null or 0.

PR #104, opened 2022-12-20
Rename com.beeper.hungry_row_id to com.beeper.hs.order

hifi:


event/events.go
```diff
@@ -140,7 +140,7 @@ type Unsigned struct {
 	RedactedBecause *Event          `json:"redacted_because,omitempty"`
 	InviteRoomState []StrippedState `json:"invite_room_state,omitempty"`
 
-	HungryRowID int64 `json:"com.beeper.hungry_row_id,omitempty"`
+	BeeperHsOrder int64 `json:"com.beeper.hs.order,omitempty"`
```
tulir, line 143:
```suggestion
	BeeperHSOrder int64 `json:"com.beeper.hs.order,omitempty"`
```

(could also do `BeeperHomeserverOrder`, but if it's an initialism it has to be in consistent case)

hifi: We seem to call it "HS" ordering on Synapse as well.

tulir (approved):

PR #107, opened 2022-12-30
Ignore invite error message from dendrite when user is already in the room

raf924:
In Dendrite, the error message when a user is invited in a room they've already joined is different from the one returned by Synapse.

See https://github.com/matrix-org/dendrite/blob/main/roomserver/internal/perform/perform_invite.go#L178


tulir (approved):
This is definitely not a good solution, but MSC3848 doesn't seem to be moving anywhere :(

PR #109, opened 2023-01-03
Add knocked rooms to sync response struct

Fizzadar:
Adds knocked rooms to sync response spec.

### DIFF
--- a/responses.go
+++ b/responses.go
@@ -262,6 +262,7 @@ type RespSyncRooms struct {
 	Leave  map[id.RoomID]*SyncLeftRoom    `json:"leave,omitempty"`
 	Join   map[id.RoomID]*SyncJoinedRoom  `json:"join,omitempty"`
 	Invite map[id.RoomID]*SyncInvitedRoom `json:"invite,omitempty"`
+	Knock  map[id.RoomID]*SyncKnockedRoom `json:"knock,omitempty"`
 }
 
 type marshalableRespSync RespSync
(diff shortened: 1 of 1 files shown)


responses.go
```diff
@@ -253,7 +253,7 @@ type RespSync struct {
 
 	DeviceLists    DeviceLists       `json:"device_lists"`
 	DeviceOTKCount OTKCount          `json:"device_one_time_keys_count"`
-	FallbackKeys   []id.KeyAlgorithm `json:"device_unused_fallback_key_types"`
+	FallbackKeys   []id.KeyAlgorithm `json:"device_unused_fallback_key_types,omitempty"`
```
tulir, line 256: That looks a bit sus

Fizzadar: It’s optional in the spec I believe, for non e2ee clients I’d guess?

tulir: It's required in the spec, although it won't exist in servers that don't support the spec version that added it. Is the omitempty needed for something?

responses.go
```diff
@@ -338,19 +339,26 @@ func (sir SyncInvitedRoom) MarshalJSON() ([]byte, error) {
...
-	Servers []string  `json:"servers"`
-}
+type (
+	RespAliasCreate  struct{}
+	RespAliasDelete  struct{}
+	RespAliasResolve struct {
+		RoomID  id.RoomID `json:"room_id"`
+		Servers []string  `json:"servers"`
+	}
+)
```
sumnerevans, line 360: does this formatting change actually do anything besides group these together?

If we are grouping `RespAlias*` together, why is `RespAliasList` not included here?

responses.go
Fizzadar: Oh my bad - doesn't show required under the sync spec but does under OTK, will remove this as was only making the change for spec compliance.

responses.go
Fizzadar: Woa! Not sure how this slipped in think this was `gofumpt` sneaking this in here let me fix that.

Fizzadar:
OK I have fixed this mess, just the knocked room addition now.

tulir (approved):

sumnerevans (approved):

PR #111, opened 2023-02-07
power levels event content: add handling for notifications key

sumnerevans:

### DIFF
--- a/event/powerlevels.go
+++ b/event/powerlevels.go
@@ -23,6 +23,8 @@ type PowerLevelsEventContent struct {
 	Events        map[string]int `json:"events,omitempty"`
 	EventsDefault int            `json:"events_default,omitempty"`
 
+	Notifications PowerLevelsEventContentNotifications `json:"notifications,omitempty"`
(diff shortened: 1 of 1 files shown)


tulir (approved):
Not sure if it needs to be a map since the spec only defines `room` as a valid key (could just be a struct), but I guess either way is fine.

event/powerlevels.go
```diff
@@ -32,6 +34,13 @@ type PowerLevelsEventContent struct {
 	HistoricalPtr *int `json:"historical,omitempty"`
 }
 
+// Notifications represents the content of the notifications key in an
+// m.room.power_levels state event content.
+// https://spec.matrix.org/v1.2/client-server-api/#mroompower_levels
+type PowerLevelsEventContentNotifications struct {
+	Room int `json:"room,omitempty"`
```
tulir, line 41: Should probably do `RoomPtr` similar to `InvitePtr` and others so that it can have a getter that defaults to 50 when omitted

PR #115, opened 2023-02-28
Add request and response types for room key backup

vurpo:


responses.go
```diff
@@ -562,3 +562,35 @@ type RespTimestampToEvent struct {
...
+
+type RespRoomKeysVersionCreate struct {
+	Version string `json:"version"`
+}
+
+type RespRoomKeysVersion struct {
+	Algorithm string                 `json:"algorithm"`
+	AuthData  map[string]interface{} `json:"auth_data"`
+	Count     int                    `json:"count"`
+	Etag      string                 `json:"etag"`
```
tulir, line 574: This should probably be `ETag`

requests.go
```diff
@@ -411,3 +411,23 @@ type ReqBeeperSplitRoom struct {
...
+}
+
+type ReqRoomKeysRoomUpdate struct {
+	Sessions map[id.SessionID]ReqRoomKeysSessionUpdate `json:"sessions"`
+}
+
+type ReqRoomKeysSessionUpdate struct {
+	FirstMessageIndex int                    `json:"first_message_index"`
+	ForwardedCount    int                    `json:"forwarded_count"`
+	IsVerified        bool                   `json:"is_verified"`
```
tulir, line 431: Are all the fields always required? Should probably have some `,omitempty`s if not

vurpo: Spec says that they are always required

tulir (approved):

PR #116, opened 2023-03-04
matrix server unix socket connection support

cyberb:
Matrix server now supports listening on a unix socket: https://github.com/matrix-org/dendrite/commit/6b1c9eafa97cb02dd902ddbf2676c709c86f9625
I would like to be able to use that feature as I am running multiple services on a small RPi like device and it removes port management. I have tested this on my device.

PR includes both client side (home server) and server side (app service) support.
I am not sure but it looks like appservice.address is not used.
Config looks like this:
```
homeserver:
  address: /var/snap/matrix/current/matrix.socket
...
appservice:
  address: /var/snap/matrix/current/whatsapp.socket
  hostname: /var/snap/matrix/current/whatsapp.socket
  port: 0

```

### DIFF
--- a/appservice/appservice.go
+++ b/appservice/appservice.go
@@ -7,10 +7,13 @@
 package appservice
 
 import (
+	"context"
 	"fmt"
+	"net"
 	"net/http"
 	"net/http/cookiejar"
 	"os"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -34,7 +37,6 @@ var OTKChannelSize = 4
 
 // Create a blank appservice instance.
 func Create() *AppService {
-	jar, _ := cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
 	defaultLogLevel := zerolog.TraceLevel
 	return &AppService{
 		LogConfig: &zeroconfig.Config{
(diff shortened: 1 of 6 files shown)


appservice/appservice.go
```diff
@@ -271,7 +301,7 @@ func (as *AppService) makeClient(userID id.UserID) *mautrix.Client {
 	client.StateStore = as.StateStore
 	client.SetAppServiceUserID = true
 	client.Log = as.Log.With().Str("as_user_id", client.UserID.String()).Logger()
-	client.Client = as.HTTPClient
+	client.Client = as.CreateHTTPClient()
```
tulir, line 304: Creating new clients for each ghost user doesn't seem like a good idea 🤔

cyberb: Right, but now it is possible to have a mix of app servers with different connection (tcp vs unix). Should I cache http clients per type and lazy create them?

cyberb: At the same time, are you worried about memory/performance of creating http.client?

tulir: The `AppService` struct has the http client as a field so it's not shared between different instances or anything like that.

Having multiple http clients means connections don't get reused properly

cyberb: ok, looked at again, my bad, fixed it, I was confused by the fact that AppService is created and only then also populated by config values from yaml, I think this is super confusing.

cyberb: @tulir is this better? I am testing it with dendrite and whatsapp bridge, but I think they do not work together well currently because of this: https://github.com/matrix-org/dendrite/issues/2902 other than this I think two-way unix socket communication is working ok.

cyberb: @tulir is it possible to make a decision on this PR?

PR #117, opened 2023-03-05
fix warning that is logged when attempting to unmarshal empty value in encryption column for public rooms

ashfame:
Running [example](https://github.com/mautrix/go/blob/master/example/main.go) code, I am seeing a lot of these warnings:

```
3:01PM WRN Failed to parse encryption config of !GtIfdsfQtQIgbQSxwJ:archlinux.org: unexpected end of JSON input component=crypto db_section=matrix_state` for public rooms
```

For public rooms, encryption column is empty and attempting to unmarshal logs a warning. This PR checks whether a json value is valid json before attempting to unmarshall it and returns `nil` if so.

### DIFF
--- a/sqlstatestore/statestore.go
+++ b/sqlstatestore/statestore.go
@@ -215,6 +215,9 @@ func (store *SQLStateStore) GetEncryptionEvent(roomID id.RoomID) *event.Encrypti
 		}
 		return nil
 	}
+	if !json.Valid(data) {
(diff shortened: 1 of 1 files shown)


tulir:
Good catch, but invalid JSON should produce warnings (or errors really, but I haven't got around to refactoring the state store to return errors), only nulls are safe to ignore.

PR #119, opened 2023-03-18
unix socket fix and test

cyberb:
Url is a pointer so was not copied properly and broke unix socket transport


cyberb:
Probably instead of copying (which is still shallow) we could remember Scheme and Path for later checks.

cyberb:
Fyi, it was working in the original PR as I have end to end tests in Syncloud matrix builds with chat screenshots :)

tulir (approved):

PR #122, opened 2023-04-02
Allow cryptohelper to be used with an in-memory store

mraerino:
This fixes a TODO comment found in the latest release.

The change seems pretty straight forward to me, but i'm not sure if `helper.client.StateStore` can always be used. Would love a double-check on that.


tulir (approved):
The plan was to figure out a better name or signature for the method, but I guess this solves it faster

PR #128, opened 2023-04-18
bridge bot: is_bot -> is_network_bot, and set to false

sumnerevans:


tulir (approved):
Might make more sense to omit it but I guess this works too

PR #131, opened 2023-06-05
Add full RO pool support for SQLite

hifi:


hifi:
Made this more generic by allowing user configured drivers to be used so our downstream test in Hungryserv can set the busy timeout to zero making unit tests non-blocking without compromising on the test scope.

Downstream PR depends on this to be merged first to make the tests pass.

hifi:
Rejigged based on our conversation.

util/dbutil/database.go
```diff
@@ -186,54 +194,87 @@ type Config struct {
...
+		if len(uriParts) == 2 {
+			var err error
+			qs, err = url.ParseQuery(uriParts[1])
+			if err != nil {
+				return nil, err
+			}
+
+			qs.Del("_txlock")
+		}
+		qs.Set("_query_only", "true")
```
tulir, line 266: The uri modification should probably only happen if the read-only pool uri is empty

util/dbutil/database.go
```diff
@@ -186,54 +194,87 @@ type Config struct {
...
-		return nil, err
-	}
-	conn, err := sql.Open(cfg.Type, cfg.URI)
+	wrappedDB, err := NewWithDialect(cfg.URI, cfg.Type)
 	if err != nil {
 		return nil, err
 	}
 	if logger == nil {
 		logger = NoopLogger
 	}
```
tulir, line 241: This can be simplified to only change `wrappedDB.Log` when `logger != nil`

util/dbutil/database.go
```diff
@@ -172,10 +174,16 @@ func NewWithDialect(uri, rawDialect string) (*Database, error) {
...
 		return nil, err
 	}
-	return NewWithDB(db, rawDialect)
+
+	wrappedDB, err := NewWithDB(db, rawDialect)
+	if err != nil {
+		return nil, err
+	}
+
+	return wrappedDB, nil
```
tulir, line 183: This change is now unnecessary

tulir (approved):

PR #132, opened 2023-06-14
Use HTTPError in checkpoint function

vurpo:


tulir (approved):
Wrapped error should be included when sending the request fails (network errors), otherwise fine

PR #133, opened 2023-06-19
properly handle http errors inside download method

Fusl:


tulir (approved):
Seems fairly sensible, thanks!

PR #137, opened 2023-06-26
logging: don't log request start, just log end

sumnerevans:

### DIFF
--- a/client.go
+++ b/client.go
@@ -14,7 +14,6 @@ import (
 	"net/url"
 	"os"
 	"strconv"
-	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -285,11 +277,15 @@ func (cli *Client) LogRequestDone(req *http.Request, resp *http.Response, handle
 	path = strings.TrimPrefix(path, "/_matrix/client")
 	evt := zerolog.Ctx(req.Context()).Debug().
 		Str("method", req.Method).
+		Str("url", req.URL.String()).
```
sumnerevans, line 280: I'm not sure if we want to include this or not. It was included in the log at the start of the request, so I thought it might be good to include here as well.

@tulir any thoughts?

tulir: Yeah should probably be here, makes `path` redundant though

client.go
```diff
@@ -1388,14 +1380,18 @@ func (cli *Client) downloadContext(ctx context.Context, mxcURL id.ContentURI) (*
 		return nil, err
 	}
 	req.Header.Set("User-Agent", cli.UserAgent+" (media downloader)")
-	cli.LogRequest(req)
+	cli.RequestStart(req)
+	downloadStart := time.Now()
 	if resp, err := cli.Client.Do(req); err != nil {
+		cli.LogRequestDone(req, resp, err, 0, time.Since(downloadStart))
```
tulir, line 1386: This call is wrong since `resp` is nil, although now I wonder if there should be some common place for logging failed requests 🤔

client.go
```diff
@@ -581,10 +589,10 @@ func (cli *Client) executeCompiledRequest(req *http.Request, retries int, backof
 	var body []byte
 	if res.StatusCode < 200 || res.StatusCode >= 300 {
 		body, err = ParseErrorResponse(req, res)
-		cli.LogRequestDone(req, res, nil, len(body), duration)
+		cli.LogRequestDone(req, res, err, nil, len(body), duration)
```
tulir, line 592: Bad HTTP status codes shouldn't be logged as errors at this layer since they might not be errors

```suggestion
		cli.LogRequestDone(req, res, nil, nil, len(body), duration)
```

tulir (approved):

PR #138, opened 2023-07-03
Retries for media fetch

vurpo:
I initially tried making use of the existing request with retry logic (the executeCompiledRequest method) but it didn't seem well suited for this kind of request and the code turned out ugly, so I made a copy of it to adapt it to this use better.

The alternative (reusing the existing method) would have looked like

```
	var resp *http.Response
	_, err := cli.MakeFullRequest(FullRequest{
		Method:  http.MethodGet,
		URL:     cli.GetDownloadURL(mxcURL),
		Context: ctx,
		Handler: func(req *http.Request, res *http.Response, _ interface{}) ([]byte, error) {
			resp = res
			return []byte{}, nil
		},
	})
	if err != nil {
		return nil, err
	}
	return resp, nil
```

instead, which is a much smaller change but I thought it looked ugly.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1412,9 +1412,6 @@ func (cli *Client) doMediaRequest(req *http.Request, retries int, backoff time.D
 	startTime := time.Now()
 	res, err := cli.Client.Do(req)
 	duration := time.Now().Sub(startTime)
-	if res != nil {
-		defer res.Body.Close()
-	}
 	if err != nil {
 		if retries > 0 {
 			return cli.doMediaRetry(req, err, retries, backoff)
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -1386,30 +1386,80 @@ func (cli *Client) DownloadContext(ctx context.Context, mxcURL id.ContentURI) (i
...
-		body, err := ParseErrorResponse(req, resp)
-		_ = resp.Body.Close()
-		cli.LogRequestDone(req, resp, err, nil, len(body), time.Since(downloadStart))
-		return nil, err
-	} else {
-		cli.LogRequestDone(req, resp, nil, nil, -1, time.Since(downloadStart))
-		return resp, nil
-	}
+	resp, err := cli.doMediaRequest(req, cli.DefaultHTTPRetries, 4*time.Second)
+	return resp, err
```
sumnerevans, line 1462:
```suggestion
	return cli.doMediaRequest(req, cli.DefaultHTTPRetries, 4*time.Second)
```

sumnerevans (commented):
I'd be interested to see what @tulir thinks about ideas to refactor this to reduce code duplication with the other retry path, but overall, it looks pretty reasonable.

tulir (approved):
Looks about right

(does it work?)

vurpo:
Works according to my testing now

PR #144, opened 2023-09-21
Add context aware funcs for all existing api functions

recht:
Most functions didn't do this, which meant that timeouts, tracing, etc didn't work correctly.


tulir:
Contexts are good in general, not sure if there's any reason to duplicate all the functions instead of just adding a context parameter 🤔

There are lots of breaking changes anyway, the library is still v0.x

recht:
I can definitely change the existing functions instead, I just didn't know if backwards compatibility was a priority. I'll update the PR to just change the existing functions since that is much cleaner.

recht:
I updated the PR to change existing functions - that required quite a lot of other changes, but on the positive side context propagation is then everywhere

PR #145, opened 2023-09-22
Add a SyncStore2 interface which supports context and error handling

recht:
This makes it possible to gracefully fail if store reads/updates fail instead of panicking.

A new interface is added in order to preserve backwards compatibility.


recht:
#144 does this now too

PR #153, opened 2024-01-10
Allow disabling automatic key fetching for Olm machine

hifi:
Many crypto operations in the Olm machine have a possible side effect of fetching keys from the server if they are missing. This may be undesired in some special cases.

To tracking which users need key fetching, CryptoStore now exposes APIs to mark and query the status.

### DIFF
--- a/crypto/encryptmegolm.go
+++ b/crypto/encryptmegolm.go
@@ -229,12 +229,16 @@ func (mach *OlmMachine) ShareGroupSession(ctx context.Context, roomID id.RoomID,
 
 	if len(fetchKeys) > 0 {
 		log.Debug().Strs("users", strishArray(fetchKeys)).Msg("Fetching missing keys")
-		for userID, devices := range mach.fetchKeys(ctx, fetchKeys, "", true) {
-			log.Debug().
-				Int("device_count", len(devices)).
-				Str("target_user_id", userID.String()).
-				Msg("Got device keys for user")
-			missingSessions[userID] = devices
+		if keys, err := mach.FetchKeys(ctx, fetchKeys, true); err != nil {
+			log.Err(err).Strs("users", strishArray(fetchKeys)).Msg("Failed to fetch missing keys")
+		} else if keys != nil {
+			for userID, devices := range keys {
+				log.Debug().
+					Int("device_count", len(devices)).
+					Str("target_user_id", userID.String()).
+					Msg("Got device keys for user")
+				missingSessions[userID] = devices
+			}
 		}
 	}
 
--- a/crypto/sql_store.go
+++ b/crypto/sql_store.go
@@ -665,12 +665,19 @@ func (store *SQLCryptoStore) PutDevice(ctx context.Context, userID id.UserID, de
 	return err
 }
 
+const trackedUserUpsertQuery = `
+INSERT INTO crypto_tracked_user (user_id, devices_outdated)
+VALUES ($1, false)
+ON CONFLICT (user_id) DO UPDATE
+	SET devices_outdated = EXCLUDED.devices_outdated
+`
+
 // PutDevices stores the device identity information for the given user ID.
 func (store *SQLCryptoStore) PutDevices(ctx context.Context, userID id.UserID, devices map[id.DeviceID]*id.Device) error {
 	return store.DB.DoTxn(ctx, nil, func(ctx context.Context) error {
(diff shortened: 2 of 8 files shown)


crypto/encryptmegolm.go
```diff
@@ -228,13 +228,21 @@ func (mach *OlmMachine) ShareGroupSession(ctx context.Context, roomID id.RoomID,
...
 	if len(fetchKeys) > 0 {
-		log.Debug().Strs("users", strishArray(fetchKeys)).Msg("Fetching missing keys")
-		for userID, devices := range mach.fetchKeys(ctx, fetchKeys, "", true) {
-			log.Debug().
-				Int("device_count", len(devices)).
-				Str("target_user_id", userID.String()).
-				Msg("Got device keys for user")
-			missingSessions[userID] = devices
+		if mach.DisableKeyFetching {
+			log.Warn().Strs("users", strishArray(fetchKeys)).Msg("Keys missing but key fetching is disabled")
```
tulir, line 232: Sending requires network requests by definition so no point in disabling extra requests here

crypto/sql_store.go
```diff
@@ -668,9 +668,9 @@ func (store *SQLCryptoStore) PutDevice(ctx context.Context, userID id.UserID, de
 // PutDevices stores the device identity information for the given user ID.
 func (store *SQLCryptoStore) PutDevices(ctx context.Context, userID id.UserID, devices map[id.DeviceID]*id.Device) error {
 	return store.DB.DoTxn(ctx, nil, func(ctx context.Context) error {
-		_, err := store.DB.Exec(ctx, "INSERT INTO crypto_tracked_user (user_id) VALUES ($1) ON CONFLICT (user_id) DO NOTHING", userID)
+		_, err := store.DB.Exec(ctx, "INSERT OR REPLACE INTO crypto_tracked_user (user_id, devices_outdated) VALUES ($1, FALSE)", userID)
```
tulir, line 671: `ON CONFLICT DO ...` is the correct cross-platform way to upsert, don't use `OR REPLACE`

crypto/sql_store.go
```diff
@@ -734,6 +734,29 @@ func (store *SQLCryptoStore) FilterTrackedUsers(ctx context.Context, users []id.
 	return dbutil.NewRowIter(rows, dbutil.ScanSingleColumn[id.UserID]).AsList()
 }
 
+// MarkTrackedUsersOutdated flags that the device list for given users are outdated.
+func (store *SQLCryptoStore) MarkTrackedUsersOutdated(ctx context.Context, users []id.UserID) error {
+	return store.DB.DoTxn(ctx, nil, func(ctx context.Context) error {
+		for _, userID := range users {
+			_, err := store.DB.Exec(ctx, "INSERT OR REPLACE INTO crypto_tracked_user (user_id, devices_outdated) VALUES ($1, TRUE)", userID)
```
tulir, line 741: This should be `UPDATE`

crypto/sql_store.go
sumnerevans, line 671: I think we need to maintain compatibility with postgres, and I don't think that this syntax works.

```suggestion
		_, err := store.DB.Exec(ctx, "INSERT INTO crypto_tracked_user (user_id, devices_outdated) VALUES ($1, FALSE) ON CONFLICT DO UPDATE SET devices_outdated=excluded.devices_outdated", userID)
```

crypto/sql_store.go
sumnerevans, line 741: and here

hifi: Actually it doesn't work the intended way if it's an UPDATE. We want to flag users as tracked and outdated at the same time if the first time we see an encrypted event from them in a sync but we don't know their devices at all we want to flag and track them to do the device fetching.

hifi: Maybe if we rename it to `MarkUsersTrackedAndOutdated` so it is explicit what it does?

crypto/sql_store.go
```diff
@@ -734,6 +741,29 @@ func (store *SQLCryptoStore) FilterTrackedUsers(ctx context.Context, users []id.
 	return dbutil.NewRowIter(rows, dbutil.ScanSingleColumn[id.UserID]).AsList()
 }
 
+// MarkTrackedUsersOutdated flags that the device list for given users are outdated.
+func (store *SQLCryptoStore) MarkTrackedUsersOutdated(ctx context.Context, users []id.UserID) error {
+	return store.DB.DoTxn(ctx, nil, func(ctx context.Context) error {
+		for _, userID := range users {
```
tulir, line 747:
```suggestion
		// TODO refactor to use a single query
		for _, userID := range users {
```

tulir (approved):

PR #154, opened 2024-01-11
client: add key backup functions

sumnerevans:


crypto/backup/encryptedsessiondata.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, err
+	}
+
+	key, macKey, iv, err := calculateEncryptionParameters(sharedSecret)
+	if err != nil {
+		return nil, err
+	}
+
+	// Verify the MAC before decrypting.
+	if !bytes.Equal(calculateCompatMAC(macKey), esd.MAC) && !bytes.Equal(calculateMAC(macKey, esd.Ciphertext), esd.MAC) {
```
tulir, line 137: No point in doing the proper MAC here since the algorithm identifier will change before it's fixed

```suggestion
	// Verify the MAC before decrypting.
	if !bytes.Equal(calculateCompatMAC(macKey), esd.MAC) {
```

crypto/keybackup.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"github.com/rs/zerolog"
+
+	"maunium.net/go/mautrix/crypto/backup"
+	"maunium.net/go/mautrix/crypto/olm"
+	"maunium.net/go/mautrix/crypto/signatures"
+	"maunium.net/go/mautrix/id"
+)
+
+func (mach *OlmMachine) DownloadAndStoreLatestKeyBackup(ctx context.Context, megolmBackupKey *backup.MegolmBackupKey) error {
```
tulir, line 17: This function should probably be split up a bit so we can have functions for downloading a specific session and maybe all sessions for a specific room

tulir (approved):

crypto/keybackup.go
sumnerevans: I'll leave that for a future refactor once we actually want to do those things.

PR #162, opened 2024-01-24
Share room keys to known devices on request

hifi:
This is currently naive and not secure. Mostly for RFC about the store interface and ideas how to invalidate devices.

Should we invalidate all shared sessions when the device identity changes on PutDevice(s) or have a separate function to invalidate a device or something completely different?

We could key the known shares with device identity and that would not require special invalidation but since the database has no foreign keys automatic invalidation is a bit hard regardless. Also as PutDevice(s) is an UPSERT we don't know if the identity changed or not unless PutDevice implicitly does GetDevice first to compare and then invalidate.


tulir:
Probably just needs to be keyed on identity key instead of device ID (or maybe both)

tulir (approved):

PR #164, opened 2024-01-25
verificationhelper: basic infrastructure

sumnerevans:
This PR replaces the old verification code that was built-in to the `OlmMachine` with a `verificationhelper` package.

Note that the functionality is not complete, but I am going to merge this so that the framework is here and we can iterate on `master` without having such a large branch.

This new package will support QR code verification in addition to SAS.


crypto/verificationhelper/qrcode.go
```diff
@@ -0,0 +1,93 @@
...
+		SharedSecret:  random.Bytes(32),
+	}
+}
+
+// NewQRCodeFromBytes parses the bytes from a QR code scan as defined in
+// [Section 11.12.2.4.1] of the Spec.
+//
+// [Section 11.12.2.4.1]: https://spec.matrix.org/v1.9/client-server-api/#qr-code-format
+func NewQRCodeFromBytes(data []byte) (*QRCode, error) {
+	data = bytes.TrimPrefix(data, []byte("QR-Code:"))
```
tulir, line 49: What's this trim for 🤔

crypto/verificationhelper/qrcode.go
```diff
@@ -0,0 +1,93 @@
...
+	SharedSecret  []byte
+}
+
+func NewQRCode(mode QRCodeMode, txnID id.VerificationTransactionID, key1, key2 [32]byte) *QRCode {
+	return &QRCode{
+		Mode:          mode,
+		TransactionID: txnID,
+		Key1:          key1,
+		Key2:          key2,
+		SharedSecret:  random.Bytes(32),
```
tulir, line 40: Do other clients use such long shared secrets? The spec says "about 8 bytes long"

crypto/verificationhelper/qrcode.go
```diff
@@ -0,0 +1,93 @@
+package verificationhelper
```
tulir, line 1: Files should have the MPL header

tulir (approved):

crypto/verificationhelper/qrcode.go
sumnerevans: For some reason I was getting that from the QR code scanning program that I was using. I guess we can just rely on it not ever being present.

PR #165, opened 2024-01-25
goolm: deprecate custom base64 functions for base64.RawStdEncoding

sumnerevans:


crypto/goolm/cipher/pickle.go
```diff
@@ -28,14 +29,13 @@ func Pickle(key, input []byte) ([]byte, error) {
 		return nil, err
 	}
 	ciphertext = append(ciphertext, mac[:pickleMACLength]...)
-	encoded := goolm.Base64Encode(ciphertext)
-	return encoded, nil
+	return []byte(base64.RawStdEncoding.EncodeToString(ciphertext)), nil
```
tulir, line 32: The main reason I didn't do this before is that it's effectively doing `return []byte(string(base64Bytes))` instead of just `return base64Bytes`

I'm not sure if the compiler is smart enough to optimize it away, might need to add exbase64 to go-util with some helper methods like `func EncodeToBytes(enc *base64.Encoding, data []byte) []byte`

sumnerevans: I was more thinking about this as a first step towards wrapping libolm/goolm in a more sane (non-base64-based) API to the Go code.

PR #168, opened 2024-01-30
Database level support for key backup versioning

hifi:


hifi:
Not going to plumb this in yet in upstream to iterate faster, if I can get downstream implementation to work with this I'll drop the draft status and we can move the whole verification logic in later.

tulir (approved):

PR #170, opened 2024-01-30
Add permission level to deny login-matrix

AndrewFerr:
as some instances may want to discourage users from sharing their Matrix access tokens.

### DIFF
--- a/bridge/commands/doublepuppet.go
+++ b/bridge/commands/doublepuppet.go
@@ -14,6 +14,7 @@ var CommandLoginMatrix = &FullHandler{
 		Description: "Enable double puppeting.",
 		Args:        "<_access token_>",
 	},
+	RequiresLogin:           true,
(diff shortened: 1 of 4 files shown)


bridge/commands/doublepuppet.go
```diff
@@ -69,7 +69,7 @@ var CommandLogoutMatrix = &FullHandler{
 		Section:     HelpSectionAuth,
 		Description: "Disable double puppeting.",
 	},
-	RequiresLogin: true,
+	RequiresMatrixPuppeting: true,
```
tulir, line 72: All these also require login

AndrewFerr:
Rebased onto the latest release to facilitate clean merging with downstream branches.

AndrewFerr:
Putting this in draft while I try writing an alternative approach that should be simpler / less impactful than making a new permission level.

PR #173, opened 2024-02-02
Allow any UI auth for uploading cross signing keys

hifi:
Fix endless loop with UI auth causing 401 when uploading keys.

### DIFF
--- a/requests.go
+++ b/requests.go
@@ -97,8 +97,9 @@ type ReqUIAuthFallback struct {
 
 type ReqUIAuthLogin struct {
 	BaseAuthData
-	User     string `json:"user"`
-	Password string `json:"password"`
+	User     string `json:"user,omitempty"`
+	Password string `json:"password,omitempty"`
+	Token    string `json:"token,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


tulir (approved):
Looks sensible, I wonder why it was hardcoded in the first place when other UIA things aren't 🤔

hifi:
Are the small adjustments okay for now? We might change it again when I refactor downstream, though.

requests.go
```diff
@@ -430,12 +431,12 @@ type ReqBeeperSplitRoom struct {
...
 type ReqRoomKeysVersionCreate struct {
 	Algorithm id.KeyBackupAlgorithm `json:"algorithm"`
-	AuthData  json.RawMessage       `json:"auth_data"`
+	AuthData  any                   `json:"auth_data"`
 }
 
 type ReqRoomKeysVersionUpdate struct {
 	Algorithm id.KeyBackupAlgorithm `json:"algorithm"`
-	AuthData  json.RawMessage       `json:"auth_data"`
+	AuthData  any                   `json:"auth_data"`
```
tulir, line 439: This breaks hungryserv so it's not fine, should probably be a type parameter

PR #174, opened 2024-02-02
Add config to disallow manual double puppeting

AndrewFerr:
as some instances may want to discourage users from sharing their Matrix access tokens.

Disabling manual double puppeting still allows automatic double puppeting for any server configured with a login shared secret.

If both manual and automatic double puppeting are enabled for a user, allow `login-matrix` to replace the server-managed Matrix token with the one given by the command.

Also add the `refresh-matrix` command to fix potential issues with automatically-managed double puppet sessions.

### DIFF
--- a/bridge/commands/doublepuppet.go
+++ b/bridge/commands/doublepuppet.go
@@ -6,15 +6,18 @@
 
 package commands
 
+import "maunium.net/go/mautrix/bridge"
+
 var CommandLoginMatrix = &FullHandler{
 	Func: fnLoginMatrix,
 	Name: "login-matrix",
 	Help: HelpMeta{
 		Section:     HelpSectionAuth,
-		Description: "Enable double puppeting.",
+		Description: "Enable double puppeting with an access token for your Matrix account.",
 		Args:        "<_access token_>",
 	},
-	RequiresLogin: true,
+	RequiresLogin:                 true,
+	RequiresManualDoublePuppeting: true,
 }
 
 func fnLoginMatrix(ce *Event) {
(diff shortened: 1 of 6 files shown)


bridge/commands/doublepuppet.go
```diff
@@ -69,15 +79,53 @@ var CommandLogoutMatrix = &FullHandler{
...
+		if intent != nil && intent.SetAppServiceUserID {
+			ce.Reply("There is no need to refresh your double puppet, as it is currently managed by the bridge.")
+			return
+		}
+		puppet, ok := puppet.(bridge.RefreshableDoublePuppet)
+		if !ok {
+			ce.Reply("The bridge does not support refreshing your double puppet.")
+			return
+		}
+		err = puppet.RefreshCustomMXID()
```
tulir, line 122: This new function seems like a lot more effort than just `SwitchCustomMXID` like it does below

Also not sure if this command should be added at all. The as_token method is the preferred way and the other methods will probably be removed in the future.

bridge/commands/handler.go
```diff
@@ -55,6 +55,11 @@ type FullHandler struct {
 	RequiresPortal bool
 	RequiresLogin  bool
 
+	RequiresMatrixLogin            bool
```
tulir, line 58: This shouldn't exist as it turns out people don't like that help output changes based on login state (mautrix/discord#18 / https://github.com/mautrix/go/commit/d1b5e5b49214355f858f2ec7b85c2b7771ed3e91)

bridge/doublepuppet.go
```diff
@@ -118,30 +124,34 @@ var (
...
-			return
-		}
-		intent.SetAppServiceUserID = true
-		if savedAccessToken != useConfigASToken {
-			var resp *mautrix.RespWhoami
-			resp, err = intent.Whoami(ctx)
-			if err == nil && resp.UserID != mxid {
-				err = ErrMismatchingMXID
+	loginSecret, hasSecret := dp.getLoginSecret(mxid)
+	if savedAccessToken == "" || savedAccessToken == useConfigASToken {
```
tulir, line 138: This check seems to have moved to the wrong place, the old code was fine

tulir (changes requested):

bridge/doublepuppet.go
AndrewFerr: The goal of this was to preserve a generated/custom token if there was one, instead of replacing it.  But the utility of that is negligible if non-as_token methods are deprecated.  I'll just revert this.

tulir:
bridgev2 has a new permission system which allows both predefined levels and specifying individual permission flags

PR #175, opened 2024-02-09
Sign cross-signing master with device key

hifi:


tulir (approved):
Looks like it might be correct

PR #177, opened 2024-02-12
Use device signing key to verify interactive verification

hifi:
Remove unnecessary base64 as well.


sumnerevans (approved):
Right, I created that `.Bytes` function for this purpose and then forgot to actually use it. I hate that the spec doesn't ever specify whether it's a signing or identity key. Thanks for fixing this up.

PR #181, opened 2024-02-16
Add functions to Admin API and there response type

grvn-ht:
Hi !!
I'm using your lib to implement a benthos input for synapse and i needed more Admin API so i implemented it in this pull request.
Tell me if it's ok for you to add theses changes.
Have a nice day,
++


responses.go
```diff
@@ -623,3 +623,50 @@ type RespRoomKeysUpdate struct {
...
+
+// RoomsBlockResponse represents the response containing wether a room is blocked or not
+type RoomsBlockResponse struct {
+    Block		bool			`json:"block"`
+    UserId		string			`json:"user_id"`
+}
+
+// https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#room-messages-api
+type RespSyncRoomMessagesAdmin struct {
+    Chunk []*event.Event `json:"chunk"`
```
tulir, line 668: No point in making a new struct if the response format is the same as the client API /messages endpoint. A type alias may make sense (but in the synapseadmin package, like the rest of the admin API response types)

responses.go
```diff
@@ -623,3 +623,50 @@ type RespRoomKeysUpdate struct {
 	Count int    `json:"count"`
 	ETag  string `json:"etag"`
 }
+
+// Room represents the structure of each room
+type RoomInfos struct {
+    RoomID             string `json:"room_id"`
```
tulir, line 629: Admin API response structs should be in the synapseadmin package, not here

synapseadmin/client.go
```diff
@@ -20,3 +22,42 @@ type Client struct {
 func (cli *Client) BuildAdminURL(path ...any) string {
 	return cli.BuildURL(mautrix.SynapseAdminURLPath(path))
 }
+
+func (cli *Client) SyncRoomWithContextAdmin(ctx context.Context, nextBatch string, roomID string, prevBatch string) (*mautrix.RespSyncRoomMessagesAdmin, error) {
+	resSyncAdmin, err := cli.FullAdminSyncRequest(ctx, ReqSyncAdmin{
+		from: nextBatch,
```
tulir, line 28: /messages is not the same as sync. Only /sync is sync. There's no need for two methods and the method signature should be similar to the client API /messages wrapper (https://pkg.go.dev/maunium.net/go/mautrix#Client.Messages)

synapseadmin/roomapi.go
```diff
@@ -0,0 +1,167 @@
...
+		URL:          reqURL,
+		ResponseJSON: &resp,
+	})
+	return &resp, err
+}
+
+// ReqDeleteRoom is the request content for Client.DeleteRoom.
+type ReqDeleteRoom struct {
+	// Room Id to delete
+	RoomId string `json:"room_id"`
```
tulir, line 60: Initialisms like `URL` and also `ID` (when short for "identifier") should be consistent in case (i.e. `id` or `ID`, never `Id`) https://go.dev/wiki/CodeReviewComments#initialisms

Also, there are some primitive types in the `id` package which should always be used when applicable https://pkg.go.dev/maunium.net/go/mautrix/id#RoomID

(both of these apply to multiple places in the code)

synapseadmin/roomapi.go
```diff
@@ -0,0 +1,167 @@
...
+	UserId id.UserID `json:"user_id"`
+	// Room Id or alias
+	Room string `json:"room"`
+}
+
+//  Make admin a Room's user based on Id
+//
+// https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#make-room-admin-api
+func (cli *Client) RoomAdmin(ctx context.Context, req ReqRoomAdmin) error {
+	reqURL := cli.BuildAdminURL("v1", "rooms", req.Room, "make_room_admin")
```
tulir, line 104: In general, path parameters should go in the function arguments and request body should go in a struct. Don't mix them unless there's a good reason to.

(this applies to a few other methods too)

synapseadmin/roomapi.go
```diff
@@ -0,0 +1,167 @@
...
+		URL:          reqURL,
+		RequestJSON: &req,
+	})
+	return err
+}
+
+// Get Room Members based on Room Id
+//
+// https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#version-2-new-version
+func (cli *Client) RoomMembers(ctx context.Context, RoomId string) (*mautrix.RoomsMembersResponse, error) {
```
tulir, line 81: In addition to the initialism note (`Id` -> `ID`), local variable and parameter names should always start with lowercase. Only exported struct fields/methods and global variables/types/functions start with uppercase.

synapseadmin/userapi.go
```diff
@@ -109,3 +109,101 @@ func (cli *Client) GetUserInfo(ctx context.Context, userID id.UserID) (resp *Res
...
+		URL:          cli.BuildAdminURL("v1", "users", userID, "override_ratelimit"),
+		ResponseJSON: &resp,
+	})
+	return
+}
+
+// DelUserRateLimit delete RateLimit from a specific user.
+//
+// https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#query-user-account
+func (cli *Client) DelUserRateLimit(ctx context.Context, userID string) (err error) {
```
tulir, line 202: `Delete` should be spelled out fully here, no reason to shorten to `Del`

The doc link in the comment also seems to be the wrong one

synapseadmin/roomapi.go
```diff
@@ -0,0 +1,167 @@
...
+   req.Limit = 100
+   return req
+}
+
+func (req *ReqListRoom) BuildQuery() map[string]string {
+	query := map[string]string{
+		"limit": strconv.Itoa(req.Limit),
+	}
+	if req.Alias != "" {
+		query["search_term"] = req.Alias
```
tulir, line 36: Why is the search_term field called `Alias`?

synapseadmin/roomapi.go
```diff
@@ -0,0 +1,167 @@
...
+   req := ReqListRoom{}
+   req.Alias = ""
+   req.From = 0
+   req.Limit = 100
+   return req
+}
+
+func (req *ReqListRoom) BuildQuery() map[string]string {
+	query := map[string]string{
+		"limit": strconv.Itoa(req.Limit),
```
tulir, line 33: Seems like it'd be easier to omit `limit` when it's 0 and get rid of the whole `NewReqListRoom` function.

tulir (changes requested):
In general totally fine with adding new admin API wrappers, but the code quality needs a bit of improvement. Some of the comments apply to more than one place (I didn't want to repeat the same comment multiple times), so please fix all similar points

grvn-ht:
Thanks for reviewing, going to fix it

grvn-ht:
Hi @tulir ,
I made requested changes,
Is everything good or are there still things to do?
Have a nice day!

synapseadmin/responses.go
```diff
@@ -0,0 +1,54 @@
...
+    StateEvents        int    		`json:"state_events"`
+    RoomType           string 		`json:"room_type"`
+}
+
+// RoomsResponse represents the response containing a list of rooms
+type RoomsResponse struct {
+    Rooms      []RoomInfos `json:"rooms"`
+    Offset     int    `json:"offset"`
+    TotalRooms int    `json:"total_rooms"`
+	NextBatch  int	  `json:"next_batch"`
```
tulir, line 38: The response also seems to have `next_token` and `prev_batch`

synapseadmin/responses.go
```diff
@@ -0,0 +1,54 @@
...
+// RoomsMembersResponse represents the response containing a list of members of a room
+type RoomsMembersResponse struct {
+    Members		[]string	`json:"members"`
+    Total		int			`json:"total"`
+}
+
+// RoomsBlockResponse represents the response containing wether a room is blocked or not
+type RoomsBlockResponse struct {
+    Block		bool		`json:"block"`
+    UserId		id.UserID	`json:"user_id"`
```
tulir, line 50: `UserId` casing is wrong (should be `UserID`)

synapseadmin/roomapi.go
```diff
@@ -0,0 +1,153 @@
...
+		URL:          reqURL,
+		ResponseJSON: &resp,
+	})
+	return resp, err
+}
+
+// ReqDeleteRoom is the request content for Client.DeleteRoom.
+type ReqDeleteRoom struct {
+	// Default to true: remove all traces from database
+	Purge bool `json:"purge"`
```
tulir, line 54: This should also have the other fields (`block`, `message`, `room_name`, `new_room_user_id`). Also, all the fields are optional, so they should have `omitempty` in the json struct tag

synapseadmin/roomapi.go
```diff
@@ -0,0 +1,153 @@
...
+		URL:          reqURL,
+		RequestJSON: &req,
+	})
+	return err
+}
+
+// Get Room Members based on Room Id
+//
+// https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#version-2-new-version
+func (cli *Client) RoomMembers(ctx context.Context, roomId id.RoomID) (RoomsMembersResponse, error) {
```
tulir, line 73: `roomId` -> `roomID`

synapseadmin/roomapi.go
```diff
@@ -0,0 +1,153 @@
...
+		URL:          reqURL,
+		ResponseJSON: &resp,
+	})
+	return resp, err
+}
+
+// ReqRoomAdmin is the request content for Client.RoomAdmin.
+type ReqRoomAdmin struct {
+	// User Id to make admin
+	userID id.UserID `json:"user_id"`
```
tulir, line 87: Struct fields need to be exported for them to be used in JSON (`userID` -> `UserID`)

synapseadmin/roomapi.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+//  Make admin a Room's user based on room Id or room alias
+//
+// https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#make-room-admin-api
+func (cli *Client) RoomAdmin(ctx context.Context, room string, req ReqRoomAdmin) error {
```
tulir, line 93: `room string` -> `roomID id.RoomID`

synapseadmin/responses.go
```diff
@@ -0,0 +1,54 @@
...
+// RoomInfos represents the structure of each room
+type RoomInfos struct {
+    RoomID             id.RoomID	`json:"room_id"`
+    Name               string		`json:"name"`
+    CanonicalAlias     string		`json:"canonical_alias"`
+    JoinedMembers      int    		`json:"joined_members"`
+    JoinedLocalMembers int    		`json:"joined_local_members"`
+    Version            string 		`json:"version"`
+    Creator            string 		`json:"creator"`
+    Encryption         string 		`json:"encryption"`
```
tulir, line 23: Some of these should have proper types:

CanonicalAlias: `id.RoomAlias`
Creator: `id.UserID`
Encryption: `id.Algorithm`

synapseadmin/responses.go
```diff
@@ -0,0 +1,54 @@
...
+    Federatable        bool   		`json:"federatable"`
+    Public             bool   		`json:"public"`
+    JoinRules          string 		`json:"join_rules"`
+    GuestAccess        string 		`json:"guest_access"`
+    HistoryVisibility  string 		`json:"history_visibility"`
+    StateEvents        int    		`json:"state_events"`
+    RoomType           string 		`json:"room_type"`
```
tulir, line 30: `event.JoinRule`, `event.GuestAccess` and `event.HistoryVisibility` as the types for the first three fields, `event.RoomType` for the last one. (`StateEvents` is fine as an int, no need to change that)

tulir (changes requested):
Looking better, but some fields still need better types and the `Messages` function has a duplicate.

synapseadmin/roomapi.go
tulir: Oh nevermind, this endpoint supports both room IDs and room aliases. Then keep the type as string, but rename the field to `roomIDOrAlias` so it's clear

grvn-ht:
Thanks for reviewing @tulir !
Everything up to date now
Except for one lint problem in crypto/olm/account.go
fatal error: olm/olm.h: No such file or directory
But I didn't played with this project's part, it should have already been there
Tell me if you see other things to change in my code!

synapseadmin/client.go
```diff
@@ -20,3 +23,43 @@ type Client struct {
...
+	if req.from != "" {
+		query["from"] = req.from
+	}
+	if req.to != "" {
+		query["to"] = req.to
+	}
+	return query
+}
+
+func (cli *Client) FullAdminSyncRequest(ctx context.Context, req ReqSyncAdmin, roomID id.RoomID) (resp RespMessagesAdmin, err error) {
```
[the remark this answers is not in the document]
grvn-ht: done, forgot to remove func on last update

grvn-ht:
need more changes or all good?

sumnerevans:
As Tulir mentioned, the linter must pass.

grvn-ht:
my local config was missing goimport ^^
now it should be good

tulir (approved):

PR #187, opened 2024-02-21
Use the encoded form of megolm session key in backup session data

hifi:
We're using the encoded presentation elsewhere as a string and this inconsistency is a footgun.

### DIFF
--- a/crypto/backup/megolmbackup.go
+++ b/crypto/backup/megolmbackup.go
@@ -35,5 +35,5 @@ type MegolmSessionData struct {
 	ForwardingKeyChain []string          `json:"forwarding_curve25519_key_chain"`
 	SenderClaimedKeys  SenderClaimedKeys `json:"sender_claimed_keys"`
 	SenderKey          id.SenderKey      `json:"sender_key"`
-	SessionKey         []byte            `json:"session_key"`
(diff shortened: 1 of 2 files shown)


crypto/backup/megolmbackup.go
```diff
@@ -35,5 +35,5 @@ type MegolmSessionData struct {
 	ForwardingKeyChain []string          `json:"forwarding_curve25519_key_chain"`
 	SenderClaimedKeys  SenderClaimedKeys `json:"sender_claimed_keys"`
 	SenderKey          id.SenderKey      `json:"sender_key"`
-	SessionKey         []byte            `json:"session_key"`
```
sumnerevans, line 38: We should keep this `[]byte` and make all places where we set it perform a base64-decode if the value is base64-encoded.

Ideally, we will eventually just everything crypto-related deal only with `[]byte`, and the libolm implementation will just base64-encode before passing to libolm.

hifi: If you want to keep it as a byte slice it should at least be a custom type because libolm returns a byte slice _of base64_ which is extremely confusing. We also have SessionKey as a string in ExportedSession so all usages should be unified regardless.

tulir (approved):

PR #188, opened 2024-03-01
Fork from v0.16.2

AndrewFerr:
This PR rewinds the fork point of this repo to v0.16.2 (instead of v0.17.0) so that this fork can be used by our mautrix-whatsapp fork, which still depends on mautrix-go v0.16.2.

The only code difference between this PR branch and the current `element-master` is the removal of the `examples/` directory, as we don't need them & keeping them up-to-date has made merges more difficult.

This PR can be merged by either a regular merge (which would preserve both versions of commit history), or a force push/reset (which would not preserve history).  Either way, tags for all prior releases will remain, and the equivalents of prior releases in this PR branch will be given `-mod-2` tags.


AndrewFerr:
Augh, this targets the wrong repo. Closing.

tulir:
I'll hopefully get around to updating whatsapp to mautrix-go v0.18 before this month's release

PR #196, opened 2024-03-12
Use a callback to validate secret response

hifi:
To properly receive and store a requested secret, we usually need to validate it against something like a public key to ensure we got the correct one.

This changes the API so that we instead use a callback to accept any incoming secret matching our request but we'll fail when we hit the specified timeout if we never receive anything that is accepted.


hifi:
The API is slightly controversial but I haven't figured out any better way to process a secret only once on receive while also accepting or declining it at the same time.

We can't use generics when it's a method of OlmMachine to seamlessly return a type that we expect the secret to be while at the same time having a processing callback that will return the validated and processed secret. If we only have a validating callback we need to return the raw secret and re-process it which removes much of the ergonomics of this.

The expected flow is somewhat like this if there can't be any fatal errors during receiving, returning an error would bubble up:
```go
var secret SomeSecretType
err = helper.Machine().GetOrRequestSecret(ctx, name, func(s string) (bool, error) {
    if data, err := base64.StdEncoding.DecodeString(s); err != nil {
        log.Warn().Err(err).Msg("Failed to decode secret, rejecting it")
        return false, nil
    } else if ok, err := validateAndUnmarshalSomeSecretType(&secret, data); err != nil {
        log.Warn().Err(err).Msg("Failed to validate and unmarshal secret, rejecting it")
        return false, nil
    } else if ok {
        // secret was validated and unmarshaled correctly, break out
        return true, nil
    }

    return false, nil
}, time.Minute)
```

Still open to better ideas, but I have none at this point other than also wrapping this in generics as a separate helper function to create a returning version using this.

tulir (approved):

PR #202, opened 2024-03-28
Assume the device list is up-to-date on key backup restore

hifi:
Fetching devices in a loop can cause request storming if there's a lot of unknown signatures for a key backup.

A client implementation should always ensure that the devices are updated from device list changed updates from sync.


tulir (approved):
Looks fairly sensible

PR #203, opened 2024-04-02
Invalidate memory cache when storing own cross-signing keys

hifi:
When another device does cross-signing reset we would incorrectly cache the old keys indefinitely.

### DIFF
--- a/crypto/cross_sign_store.go
+++ b/crypto/cross_sign_store.go
@@ -96,5 +96,12 @@ func (mach *OlmMachine) storeCrossSigningKeys(ctx context.Context, crossSigningK
 				}
 			}
 		}
+
+		// Clear internal cache so that it refreshes from crypto store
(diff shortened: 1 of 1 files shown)


crypto/cross_sign_store.go
```diff
@@ -96,5 +96,11 @@ func (mach *OlmMachine) storeCrossSigningKeys(ctx context.Context, crossSigningK
 				}
 			}
 		}
+
+		// Clear internal cache so that it refreshes from crypto store
+		if userID == mach.Client.UserID && mach.crossSigningPubkeys != nil {
+			log.Debug().Err(err).Msg("Resetting internal cross-signing key cache")
+			mach.crossSigningPubkeys = nil
```
tulir, line 103:
```suggestion
			mach.crossSigningPubkeys = nil
			mach.crossSigningPubkeysFetched = false
```

tulir (approved):

PR #216, opened 2024-05-14
Pass error to `Client.ResponseHook`

Fizzadar:
The current `ResponseHook` doesn't pass the error through.

### DIFF
--- a/client.go
+++ b/client.go
@@ -65,7 +65,7 @@ type Client struct {
 	Log zerolog.Logger
 
 	RequestHook  func(req *http.Request)
-	ResponseHook func(req *http.Request, resp *http.Response, duration time.Duration)
+	ResponseHook func(req *http.Request, resp *http.Response, err error, duration time.Duration)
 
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -64,8 +64,9 @@ type Client struct {
 
 	Log zerolog.Logger
 
-	RequestHook  func(req *http.Request)
-	ResponseHook func(req *http.Request, resp *http.Response, duration time.Duration)
+	RequestHook       func(req *http.Request)
+	ResponseHook      func(req *http.Request, resp *http.Response, duration time.Duration)
+	ResponseErrorHook func(req *http.Request, resp *http.Response, err error, duration time.Duration)
```
tulir, line 69: The error parameter should probably just be added to ResponseHook

Fizzadar: Do we care about breaking the API?

tulir: Not really, I don't even remember what the hooks were added for

PR #218, opened 2024-05-14
crypto/sql_store: add logging on PutGroupSession

sumnerevans:


tulir (approved):

adamvy (commented):
Seems correct to me but I don't have time/context to review all the calling code in detail.  A slow rollout in bridges or beeper android would be a good test.

tulir (approved):

PR #220, opened 2024-05-20
verificationhelper: fixes for when we are generating the QR code

sumnerevans:
This PR fixes issues when the device generating the QR code does not trust the master key.


tulir (approved):
Looks about right

PR #224, opened 2024-05-21
Add `UpdateRequestOnRetry` client hook

Fizzadar:
Enables modifying the request object between retries, eg. to switch contexts after cancel.


tulir (approved):
Might also make sense to pass the `cause` so it can be updated when the context is canceled, but I guess it also works to always update

PR #225, opened 2024-05-22
verificationhelper: add E2E tests and fix QR code generation and verification logic

sumnerevans:
This PR adds end-to-end tests for the verificationhelper module. In the future, we might want to extract some of the test code such as the mock server into a separate test module if we need to E2E test anything else.

The E2E tests primarily test the happy-path, but there are some tests for the error cases as well. There are many happy paths because there are many different device combinations (scanning device vs scanned device, trust/doesn't trust the master key, etc.). I've tried to exhaustively test the happy paths.

This PR also fixes some verification bugs. Namely, it makes it so that the signing key is used in the QR code that gets generated instead of the identity key. This lines up with [what the Rust SDK does](https://github.com/matrix-org/matrix-rust-sdk/blob/1dc370942b641437a885ebacad81340e2e0d7239/crates/matrix-sdk-crypto/src/verification/qrcode.rs#L621).

I asked in [#e2e:matrix.org](https://matrix.to/#/#e2e:matrix.org) and @uhoreg responded:

> It should be the ed25519 key.  There was probably a bit of a terminology mixup in the MSC.  But all verification methods verify the ed25519 key.  In theory, devices should be able to change their curve25519 key, as long as the ed25519 key stays the same, though I don't think anyone has ever actually tried that, and I don't know what would happen if someone did. (I suspect that we would see lots of exciting errors)

It was not clear from the spec, so I'll probably create a spec clarification PR at some point to make it not so confusing to implement in the future.

### DIFF
--- a/crypto/verificationhelper/callbacks_test.go
+++ b/crypto/verificationhelper/callbacks_test.go
@@ -1,0 +1,136 @@
+// Copyright (c) 2024 Sumner Evans
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package verificationhelper_test
+
+import (
+	"context"
+
+	"maunium.net/go/mautrix/crypto/verificationhelper"
+	"maunium.net/go/mautrix/event"
+	"maunium.net/go/mautrix/id"
+)
+
+type MockVerificationCallbacks interface {
+	GetRequestedVerifications() map[id.UserID][]id.VerificationTransactionID
+	GetScanQRCodeTransactions() []id.VerificationTransactionID
+	GetQRCodeShown(id.VerificationTransactionID) *verificationhelper.QRCode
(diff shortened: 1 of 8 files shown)


crypto/cross_sign_ssss.go
```diff
@@ -100,6 +100,12 @@ func (mach *OlmMachine) GenerateAndUploadCrossSigningKeys(ctx context.Context, u
 		return "", nil, fmt.Errorf("failed to publish cross-signing keys: %w", err)
 	}
 
+	// Trust the master key
+	err = mach.SignOwnMasterKey(ctx)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to sign own master key: %w", err)
+	}
+
```
sumnerevans, line 108: @tulir I think this is the only part of this PR that I'm not quite sure if it's the right thing to do. I think it is, but I'm not 100% sure.

sumnerevans: (Discussed this out of band and we agreed that this is correct.)

PR #226, opened 2024-05-22
Fix `crypto_secrets` table schema

tulir:


crypto/sql_store_upgrade/15-fix-secrets.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE TABLE crypto_secrets_new (
+	account_id TEXT  NOT NULL,
+	name       TEXT  NOT NULL,
+	secret     bytea NOT NULL,
+
+	PRIMARY KEY (account_id, name)
+);
+
+INSERT INTO crypto_secrets_new (account_id, name, secret)
+SELECT '', name, secret
```
sumnerevans, line 11: This is more of a question for my curiosity: is the `account_id` always guaranteed to be empty if there's a single account?

sumnerevans (approved):

crypto/sql_store_upgrade/15-fix-secrets.sql
tulir: Need to confirm that before merging, `crypto_secrets` is only used in beeper android and I'm hoping the sdk always uses an empty string. If it's not always the same, `UPDATE crypto_secrets SET account_id=(SELECT account_id FROM crypto_account LIMIT 1);` will hopefully do it

tulir:
Merged as a part of #231

PR #229, opened 2024-05-23
cross signing: don't require master private key to sign master public key

sumnerevans:


tulir (approved):
Looks sensible. This change will make it fetch the public keys if they're not cached, but maybe that's fine?

PR #238, opened 2024-05-29
(*Client).Download: return entire response instead of just body

sumnerevans:


client.go
```diff
@@ -1403,12 +1403,12 @@ func (cli *Client) GetDownloadURL(mxcURL id.ContentURI) string {
...
 }
 
-func (cli *Client) Download(ctx context.Context, mxcURL id.ContentURI) (io.ReadCloser, error) {
+func (cli *Client) Download(ctx context.Context, mxcURL id.ContentURI) (*http.Response, error) {
 	resp, err := cli.download(ctx, mxcURL)
 	if err != nil {
 		return nil, err
 	}
-	return resp.Body, nil
+	return resp, nil
```
tulir, line 1411: No point in having this proxy function anymore, can just rename `download` to `Download`

PR #239, opened 2024-05-30
(*Client).SendReaction: allow specifying extra content for the send request

sumnerevans:
Useful for specifying the transaction ID manually


tulir:
These functions don't allow extra parameters because SendMessageEvent should be used directly for more complicated use cases

PR #245, opened 2024-06-19
bridgev2/directmedia: add configurable media ID prefix

sumnerevans:

### DIFF
--- a/bridgev2/bridgeconfig/config.go
+++ b/bridgev2/bridgeconfig/config.go
@@ -50,7 +50,8 @@ type ProvisioningConfig struct {
 }
 
 type DirectMediaConfig struct {
-	Enabled                bool `yaml:"enabled"`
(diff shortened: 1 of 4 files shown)


bridgev2/bridgeconfig/config.go
```diff
@@ -50,7 +50,8 @@ type ProvisioningConfig struct {
 }
 
 type DirectMediaConfig struct {
-	Enabled                bool `yaml:"enabled"`
+	Enabled                bool   `yaml:"enabled"`
+	MediaIDPrefix          string `yaml:"media_id_prefix" json:"media_id_prefix"`
```
tulir, line 54: Bridge configs are yaml only so this doesn't need the json tag

```suggestion
	MediaIDPrefix          string `yaml:"media_id_prefix"`
```

tulir (approved):
Looks about right

PR #250, opened 2024-07-02
event/powerlevels: use 0 as default required level for invite

rudis:
Hi,

I checked the other power levels and they matched the spec.

Best,
Simon


tulir (approved):
Oh yeah, this used to be wrong in the spec, I guess it was never fixed here after the spec was corrected https://github.com/matrix-org/matrix-spec/pull/1021

PR #253, opened 2024-07-09
draft: verificationhelper: Don't send networks requests in a lock

adamvy:
Also expose optional verification callback interfaces so client code can assert conformance.


adamvy:
actually this doesn't loook right, the code in cancelVerificationTxn mutates the map, so it probably should still be locked there.

but it feels wrong to send the to_device event inside this lock.  I will revisit

PR #256, opened 2024-07-17
Add MSC4144 per message profile types

hifi:


event/profile.go
```diff
@@ -0,0 +1,10 @@
+package event
+
+import "maunium.net/go/mautrix/id"
+
+type BeeperPerMessageProfile struct {
+	ID          string               `json:"id"`
+	Displayname string               `json:"displayname,omitempty"`
+	AvatarURL   *id.ContentURIString `json:"avatar_url,omitempty"`
```
tulir, line 8: This doesn't need to be a pointer, it's just a typed string

hifi: The MSC specifies that `avatar_url` being an empty string indicates the avatar being reset rather than omitted and there's no other way to do that otherwise.

tulir: oh right that was a thing 🤔

tulir (approved):

PR #259, opened 2024-07-18
keysharing: Add user_id to room key request logs

adamvy:


tulir:
It's supposed to be added by the caller https://github.com/mautrix/go/blob/main/crypto/machine.go#L395

PR #264, opened 2024-08-05
bridgev2/database: allow querying ghosts by metadata

sumnerevans:

### DIFF
--- a/bridgev2/database/ghost.go
+++ b/bridgev2/database/ghost.go
@@ -44,8 +44,9 @@ const (
 		       name_set, avatar_set, contact_info_set, is_bot, identifiers, metadata
 		FROM ghost
 	`
-	getGhostByIDQuery = getGhostBaseQuery + `WHERE bridge_id=$1 AND id=$2`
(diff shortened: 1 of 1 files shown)


bridgev2/database/ghost.go
```diff
@@ -44,8 +44,9 @@ const (
 		       name_set, avatar_set, contact_info_set, is_bot, identifiers, metadata
 		FROM ghost
 	`
-	getGhostByIDQuery = getGhostBaseQuery + `WHERE bridge_id=$1 AND id=$2`
-	insertGhostQuery  = `
+	getGhostByIDQuery       = getGhostBaseQuery + `WHERE bridge_id=$1 AND id=$2`
+	getGhostByMetadataQuery = getGhostBaseQuery + `WHERE bridge_id=$1 AND json_extract(metadata, $2)=$3`
```
tulir, line 48: Bridges have to be compatible with postgres, so the proper syntax is `metadata->>$2`, but note that paths won't be supported

PR #265, opened 2024-08-08
v2: matrix membership handling

maltee1:
Works with my local branch of mautrix-signal.

### DIFF
--- a/bridgev2/messagestatus.go
+++ b/bridgev2/messagestatus.go
@@ -36,11 +36,11 @@ var (
 	ErrUnexpectedParsedContentType     error = WrapErrorInStatus(errors.New("unexpected parsed content type")).WithErrorAsMessage().WithIsCertain(true).WithSendNotice(true)
 	ErrDatabaseError                   error = WrapErrorInStatus(errors.New("database error")).WithMessage("internal database error").WithIsCertain(true).WithSendNotice(true)
 	ErrTargetMessageNotFound           error = WrapErrorInStatus(errors.New("target message not found")).WithErrorAsMessage().WithIsCertain(true).WithSendNotice(false)
-
-	ErrUnsupportedMessageType error = WrapErrorInStatus(errors.New("unsupported message type")).WithErrorAsMessage().WithIsCertain(true).WithSendNotice(true)
-	ErrMediaDownloadFailed    error = WrapErrorInStatus(errors.New("failed to download media")).WithMessage("failed to download media").WithIsCertain(true).WithSendNotice(true)
-	ErrMediaReuploadFailed    error = WrapErrorInStatus(errors.New("failed to reupload media")).WithMessage("failed to reupload media").WithIsCertain(true).WithSendNotice(true)
-	ErrMediaConvertFailed     error = WrapErrorInStatus(errors.New("failed to convert media")).WithMessage("failed to convert media").WithIsCertain(true).WithSendNotice(true)
+	ErrUnsupportedMessageType          error = WrapErrorInStatus(errors.New("unsupported message type")).WithErrorAsMessage().WithIsCertain(true).WithSendNotice(true)
+	ErrMediaDownloadFailed             error = WrapErrorInStatus(errors.New("failed to download media")).WithMessage("failed to download media").WithIsCertain(true).WithSendNotice(true)
+	ErrMediaReuploadFailed             error = WrapErrorInStatus(errors.New("failed to reupload media")).WithMessage("failed to reupload media").WithIsCertain(true).WithSendNotice(true)
+	ErrMediaConvertFailed              error = WrapErrorInStatus(errors.New("failed to convert media")).WithMessage("failed to convert media").WithIsCertain(true).WithSendNotice(true)
+	ErrMembershipNotSupported          error = WrapErrorInStatus(errors.New("this bridge does not support changing group membership")).WithIsCertain(true).WithErrorAsMessage().WithSendNotice(true)
 )
 
 type MessageStatusEventInfo struct {
--- a/bridgev2/messagestatus.go
+++ b/bridgev2/messagestatus.go
@@ -40,7 +40,7 @@ var (
 	ErrMediaDownloadFailed             error = WrapErrorInStatus(errors.New("failed to download media")).WithMessage("failed to download media").WithIsCertain(true).WithSendNotice(true)
 	ErrMediaReuploadFailed             error = WrapErrorInStatus(errors.New("failed to reupload media")).WithMessage("failed to reupload media").WithIsCertain(true).WithSendNotice(true)
 	ErrMediaConvertFailed              error = WrapErrorInStatus(errors.New("failed to convert media")).WithMessage("failed to convert media").WithIsCertain(true).WithSendNotice(true)
-	ErrMembershipNotSupported          error = WrapErrorInStatus(errors.New("this bridge does not support changing group membership")).WithIsCertain(true).WithErrorAsMessage().WithSendNotice(true)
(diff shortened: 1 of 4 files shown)


bridgev2/networkinterface.go
```diff
@@ -593,6 +593,52 @@ type GroupCreatingNetworkAPI interface {
 	CreateGroup(ctx context.Context, name string, users ...networkid.UserID) (*CreateChatResponse, error)
 }
 
+type MembershipChangeType string
+
+const (
+	AcceptInvite  MembershipChangeType = "invite,join,1"
```
tulir, line 599: This should probably just be a real struct instead of a struct in a string. It can still have variables for each type (can't do constants though). Non-pointer structs can be compared using `==` too

maltee1: that's what I was initially going to do, but isn't mutability of those structs a problem? Those should really not change...

tulir: Non-pointer structs are always copied, so there's no way to accidentally mutate the original variable

bridgev2/portal.go
```diff
@@ -3244,3 +3247,66 @@ func (portal *Portal) SetRelay(ctx context.Context, relay *UserLogin) error {
 	}
 	return nil
 }
+
+func handleMatrixMembership(
+	portal *Portal,
+	ctx context.Context,
+	sender *UserLogin,
+	origSender *OrigSender,
+	evt *event.Event) {
```
tulir, line 3256: The other handlers aren't methods because struct methods can't have their own type parameters, but this one doesn't have any type parameters

```suggestion
func (portal *Portal) handleMatrixMembership(
	ctx context.Context,
	sender *UserLogin,
	origSender *OrigSender,
	evt *event.Event,
) {
```

bridgev2/portal.go
```diff
@@ -493,6 +494,8 @@ func (portal *Portal) handleMatrixEvent(sender *User, evt *event.Event) {
 		handleMatrixAccountData(portal, ctx, login, evt, TagHandlingNetworkAPI.HandleRoomTag)
 	case event.AccountDataBeeperMute:
 		handleMatrixAccountData(portal, ctx, login, evt, MuteHandlingNetworkAPI.HandleMute)
+	case event.StateMember:
+		handleMatrixMembership(portal, ctx, login, origSender, evt)
```
tulir, line 498:
```suggestion
		portal.handleMatrixMembership(ctx, login, origSender, evt)
```

bridgev2/networkinterface.go
```diff
@@ -593,6 +593,44 @@ type GroupCreatingNetworkAPI interface {
...
+	RevokeInvite  = MembershipChangeType{From: event.MembershipInvite, To: event.MembershipLeave}
+	RejectInvite  = MembershipChangeType{From: event.MembershipInvite, To: event.MembershipLeave, IsSelf: true}
+	BanInvited    = MembershipChangeType{From: event.MembershipInvite, To: event.MembershipBan}
+	ProfileChange = MembershipChangeType{From: event.MembershipJoin, To: event.MembershipJoin, IsSelf: true}
+	Leave         = MembershipChangeType{From: event.MembershipJoin, To: event.MembershipLeave, IsSelf: true}
+	Kick          = MembershipChangeType{From: event.MembershipJoin, To: event.MembershipLeave}
+	BanJoined     = MembershipChangeType{From: event.MembershipJoin, To: event.MembershipBan}
+	Invite        = MembershipChangeType{From: event.MembershipLeave, To: event.MembershipInvite}
+	Join          = MembershipChangeType{From: event.MembershipLeave, To: event.MembershipJoin}
+	Ban           = MembershipChangeType{From: event.MembershipLeave, To: event.MembershipBan}
```
tulir, line 613: This should have the from membership to be consistent with the others (it's not even the most common case, so having a generic name doesn't make sense)

```suggestion
	BanLeft       = MembershipChangeType{From: event.MembershipLeave, To: event.MembershipBan}
```

bridgev2/messagestatus.go
```diff
@@ -36,11 +36,11 @@ var (
...
-	ErrMediaConvertFailed     error = WrapErrorInStatus(errors.New("failed to convert media")).WithMessage("failed to convert media").WithIsCertain(true).WithSendNotice(true)
+	ErrUnsupportedMessageType          error = WrapErrorInStatus(errors.New("unsupported message type")).WithErrorAsMessage().WithIsCertain(true).WithSendNotice(true)
+	ErrMediaDownloadFailed             error = WrapErrorInStatus(errors.New("failed to download media")).WithMessage("failed to download media").WithIsCertain(true).WithSendNotice(true)
+	ErrMediaReuploadFailed             error = WrapErrorInStatus(errors.New("failed to reupload media")).WithMessage("failed to reupload media").WithIsCertain(true).WithSendNotice(true)
+	ErrMediaConvertFailed              error = WrapErrorInStatus(errors.New("failed to convert media")).WithMessage("failed to convert media").WithIsCertain(true).WithSendNotice(true)
+	ErrMembershipNotSupported          error = WrapErrorInStatus(errors.New("this bridge does not support changing group membership")).WithIsCertain(true).WithErrorAsMessage().WithSendNotice(true)
```
tulir, line 43: This definitely shouldn't send notices in relayed groups, it might make sense when not relaying, but that's slightly more complicated to do

```suggestion
	ErrMembershipNotSupported          error = WrapErrorInStatus(errors.New("this bridge does not support changing group membership")).WithIsCertain(true).WithErrorAsMessage().WithSendNotice(false)
```

bridgev2/portal.go
```diff
@@ -3244,3 +3247,66 @@ func (portal *Portal) SetRelay(ctx context.Context, relay *UserLogin) error {
...
+		portal.sendErrorStatus(ctx, evt, ErrMembershipNotSupported)
+		return
+	}
+	targetMXID := id.UserID(*evt.StateKey)
+	isSelf := sender.User.MXID == targetMXID
+	var err error
+	var targetUserLogin *UserLogin
+	targetGhost, err := portal.Bridge.GetGhostByMXID(ctx, targetMXID)
+	if err != nil {
+		log.Err(err).Stringer("mxid", targetMXID).Msg("failed getting target ghost")
```
tulir, line 3268:
```suggestion
		log.Err(err).Stringer("mxid", targetMXID).Msg("Failed to get target ghost")
```

bridgev2/portal.go
```diff
@@ -3244,3 +3247,66 @@ func (portal *Portal) SetRelay(ctx context.Context, relay *UserLogin) error {
...
+		return
+	}
+	if targetGhost == nil {
+		targetUser, err := portal.Bridge.GetUserByMXID(ctx, targetMXID)
+		if err != nil {
+			log.Err(err).Stringer("mxid", targetMXID).Msg("failed getting target user")
```
tulir, line 3274:
```suggestion
			log.Err(err).Stringer("mxid", targetMXID).Msg("Failed to get target user")
```

bridgev2/portal.go
```diff
@@ -3244,3 +3247,66 @@ func (portal *Portal) SetRelay(ctx context.Context, relay *UserLogin) error {
...
+			return
+		}
+		targetUserLogin, _, err = portal.FindPreferredLogin(ctx, targetUser, false)
+		if err != nil {
+			log.Err(err).Stringer("mxid", targetMXID).Msg("failed getting target user login")
```
tulir, line 3279:
```suggestion
			log.Err(err).Stringer("mxid", targetMXID).Msg("Failed to get target user login")
```

tulir (approved):

bridgev2/portal.go
```diff
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
```
tulir, line 20: Oops, missed this, global logs are forbidden, make sure to use `log := zerolog.Ctx(ctx)` where necessary

PR #267, opened 2024-08-15
Beeper extension for client side media dedup

hifi:
The client can send a unique id (like a hash) of a file it intends to upload in the create request. If the server already has the file it will return a "completed" response.

The unique id is meant to be opaque for the server. For privacy reasons it is recommended not to use the raw hash of a file.  The returned MXC should be stable for the same unique id for the same user but is not guaranteed.

If the file has been created but not uploaded the response will not have a completed timestamp which allows the client to retry sending the file. If the upload has already been completed the upload URL will be empty.

It is possible for multiple clients to send a create request simultaneously with the same unique id and upload the file at the same time. It is also possible for the server to forget the unique id and allow reuploading the same file again returning a new MXC.

### DIFF
--- a/responses.go
+++ b/responses.go
@@ -111,10 +111,14 @@ type RespMediaUpload struct {
 
 // RespCreateMXC is the JSON response for https://spec.matrix.org/v1.7/client-server-api/#post_matrixmediav1create
 type RespCreateMXC struct {
-	ContentURI      id.ContentURI `json:"content_uri"`
-	UnusedExpiresAt int           `json:"unused_expires_at,omitempty"`
+	ContentURI      id.ContentURI      `json:"content_uri"`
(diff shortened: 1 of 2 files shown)


responses.go
```diff
@@ -115,6 +115,10 @@ type RespCreateMXC struct {
 	UnusedExpiresAt int           `json:"unused_expires_at,omitempty"`
 
 	UnstableUploadURL string `json:"com.beeper.msc3870.upload_url,omitempty"`
+
+	// Beeper extensions for uploading unique media only once
+	BeeperUniqueID    string `json:"com.beeper.unique_id,omitempty"`
+	BeeperCompletedAt int    `json:"com.beeper.completed_at,omitempty"`
```
tulir, line 121: Should this be a `jsontime.UnixMilli` or something like that?

tulir (approved):

PR #268, opened 2024-08-19
Don't retry client requests with the context is cancelled

bradtgmurray:


client.go
```diff
@@ -613,6 +613,11 @@ func (cli *Client) executeCompiledRequest(req *http.Request, retries int, backof
 		defer res.Body.Close()
 	}
 	if err != nil {
+		if errors.Is(err, context.Canceled) {
+			log := zerolog.Ctx(req.Context())
+			log.Err(err).Msg("context cancelled, don't retry")
+			return nil, nil, err
+		}
 		if retries > 0 {
```
tulir, line 621:
```suggestion
		if retries > 0 && !errors.Is(err, context.Canceled) {
```

tulir (approved):

PR #271, opened 2024-08-21
bridgev2/userlogin: add function to get all logins for user

sumnerevans:


tulir:
`User.GetCachedUserLogins` already exists

sumnerevans:
what if they aren't cached?

tulir:
They are cached (`loadUser` calls `unlockedLoadUserLoginsByMXID`), but maybe the method should be renamed to `GetUserLogins` to be clearer 🤔

PR #272, opened 2024-08-22
ci: add test job

sumnerevans:

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -29,6 +29,30 @@ jobs:
       - name: Lint
         run: pre-commit run -a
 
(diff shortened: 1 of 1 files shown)


sumnerevans:
I'm getting a weird test failure locally, but not in CI for some reason:

image

I thought that this might fix it, but it doesn't appear to.

tulir:
The `build` job already runs tests (it even has separate jobs for goolm and libolm)

That test failure looks like it might be caused by a weird DNS server or something

sumnerevans:
This is what I get for late-night coding lol. I'm actually just blind.

PR #278, opened 2024-08-31
Add refresh token field to login request

Fizzadar:


tulir:
What's this for 🤔

Refresh tokens aren't supported

Fizzadar:
Making it match the register req spec! Also need it for Babbleserv though can wrap the struct as workaround.

tulir:
Oh yeah I guess ReqRegister already has it. Not sure if it's worth adding refresh token support anywhere right now though since auth will change with OIDC anyway.

mautrix-go clients will never send that field (so it will always be omitted), and servers can just ignore the field if they don't support refresh tokens. The field just says whether the client supports it, I'm pretty sure the server isn't required to serve a refresh token even if the client wants it

tulir (approved):

Fizzadar:
> Oh yeah I guess ReqRegister already has it. Not sure if it's worth adding refresh token support anywhere right now though since auth will change with OIDC anyway.

Ah yeah that's a good point. I'll merge because it makes login:register structs complete, but agree not needed long term.

PR #281, opened 2024-09-01
client,bridgev2: allow extended profiles for non-beeper servers using MSC4133

tulir:


client.go
```diff
@@ -1023,7 +1024,63 @@ func (cli *Client) SetAvatarURL(ctx context.Context, url id.ContentURI) (err err
...
+	return cli.setOrUpdateProfile(ctx, data, http.MethodPatch)
+}
+
+func (cli *Client) setOrUpdateProfile(ctx context.Context, data any, method string) (err error) {
+	if cli.SpecVersions.Supports(FeatureExtendedProfiles) || cli.SpecVersions.Supports(BeeperFeatureArbitraryProfileMeta) {
+		urlPath := cli.BuildClientURL("v3", "profile", cli.UserID)
+		_, err = cli.MakeRequest(ctx, method, urlPath, data, nil)
+	} else if cli.SpecVersions.Supports(UnstableFeatureExtendedProfiles) {
+		urlPath := cli.BuildClientURL("unstable", "uk.tcpip.msc4133", "profile", cli.UserID)
+		_, err = cli.MakeRequest(ctx, method, urlPath, data, nil)
```
sumnerevans, line 1049: This error doesn't look returned. Did you want to do a naked return on the last line of the function?

tulir: Oops yeah it shouldn't be nil at the end

tulir:
MSC4133 removed the PUT and PATCH endpoints, so this no longer applies

PR #283, opened 2024-09-06
event: add encrypted file info for m.room.avatar

hifi:


hifi:
Downstream code is happy with this field.

event/state.go
```diff
@@ -26,8 +26,9 @@ type RoomNameEventContent struct {
 // RoomAvatarEventContent represents the content of a m.room.avatar state event.
 // https://spec.matrix.org/v1.2/client-server-api/#mroomavatar
 type RoomAvatarEventContent struct {
-	URL  id.ContentURIString `json:"url"`
-	Info *FileInfo           `json:"info,omitempty"`
+	URL         id.ContentURIString `json:"url"`
```
tulir, line 29: Should probably add an omitempty here (the field isn't required in the current spec)

```suggestion
	URL         id.ContentURIString `json:"url,omitempty"`
```

hifi: Made that change while at it.

tulir (approved):

PR #294, opened 2024-10-16
Add matrix 1.3 api feature to refresh access token

dalees:
* Refresh tokens are returned by the Matrix homeserver when ReqLogin.RefreshToken is set.
* Add new functions to client for Refresh() which uses a previously stored refresh token to update the existing access token once it expires.
* Implements a fudge factor on the expiry duration so we refresh well before the access token expires.

### DIFF
--- a/client.go
+++ b/client.go
@@ -79,6 +79,8 @@ type Client struct {
 	UserID        id.UserID    // The user ID of the client. Used for forming HTTP paths which use the client's user ID.
 	DeviceID      id.DeviceID  // The device ID of the client.
 	AccessToken   string       // The access_token for the client.
+	RefreshToken  string       // Token to be used for refreshing AccessToken.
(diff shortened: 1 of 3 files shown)


dalees:
Happy for suggestions of changes to the interface and naming of `Refresh` and `refreshInner` functions - it made sense to me to provide some utility that could be called frequently without knowledge of the refresh time internals.

dalees:
Hi @tulir - do you have any feedback on this PR?

This affects our use of the library when using a server that has implemented token expiry, and helps with the issue (though it still requires regular messages to be sent, otherwise the refresh token expires too!)

tulir:
Probably won't bother with this since refresh tokens aren't required and the entire auth system is about to be replaced. I can add the refresh token to the login response though to ensure it can be implemented outside the library (it was already in RespRegister, not sure why it was missed for login)

PR #308, opened 2024-10-30
event: add BeeperHSOrderString to unsigned

smweber:


event/events.go
```diff
@@ -144,13 +144,14 @@ type Unsigned struct {
 	RedactedBecause *Event          `json:"redacted_because,omitempty"`
 	InviteRoomState []StrippedState `json:"invite_room_state,omitempty"`
 
-	BeeperHSOrder    int64 `json:"com.beeper.hs.order,omitempty"`
-	BeeperHSSuborder int64 `json:"com.beeper.hs.suborder,omitempty"`
-	BeeperFromBackup bool  `json:"com.beeper.from_backup,omitempty"`
+	BeeperHSOrder       int64  `json:"com.beeper.hs.order,omitempty"`
+	BeeperHSSuborder    int64  `json:"com.beeper.hs.suborder,omitempty"`
+	BeeperHSOrderString string `json:"com.beeper.hs.order_string,omitempty"`
```
tulir, line 149: btw if we always store a int64 in the string, there's a magic `,string` struct tag (similar to omitempty) that magically parses and stringifies the int at json marshaling time

```suggestion
	BeeperHSOrderString int64 `json:"com.beeper.hs.order_string,omitempty,string"`
```

tulir (approved):

event/events.go
smweber: My plan is to take advantage of the lack of size limits on the string and store the full order and subOrder in it, separated by `.`, so they can both be parsed out by local bridge manager to calculate a new hs.order for a fresh bridge, but also maintain the ability to be parsed and sorted properly.

PR #331, opened 2024-12-16
crypto/decryptolm: pass context through when unwedging devices

Fizzadar:


Fizzadar:
Yeah... does feel somewhat weird but I can't see any other way to cancel in-flight olm unwedging which we need when backgrounding apps..

tulir:
Storing a background context in the OlmMachine struct might be fine?

Fizzadar:
Switched to a public background context field that can be set, seems less worse 🙃

tulir (approved):

PR #332, opened 2024-12-17
crypto/verificationhelper: add cross-signing to SAS verification path

SpiritCroc:
When verifying a new device via SAS / emoji verification, we never cross-sign the new device, which thus never fully finishes verification.

I copied the relevant `SignOwnDevice()` call from the QR verification path, tested and confirmed it works. Review appreciated from our Go and encryption experts if this is the right way to do it.

### DIFF
--- a/crypto/verificationhelper/sas.go
+++ b/crypto/verificationhelper/sas.go
@@ -13,6 +13,7 @@ import (
 	"crypto/hmac"
 	"crypto/rand"
 	"crypto/sha256"
+	"crypto/subtle"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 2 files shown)


crypto/verificationhelper/sas.go
```diff
@@ -666,6 +666,14 @@ func (vh *VerificationHelper) onVerificationMAC(ctx context.Context, txn Verific
 				vh.cancelVerificationTxn(ctx, txn, event.VerificationCancelCodeUser, "failed to update device trust state after verifying: %w", err)
 				return
 			}
+
+			// Cross-sign their device with the self-signing key
+			if vh.mach.CrossSigningKeys != nil {
```
tulir, line 671: This probably needs a check to make sure the user being verified is the same as the current user 🤔

SpiritCroc: ok, I can look into that. Should we also add a `SignUser()` call in case it's not the same, like currently done in `ConfirmQRCodeScanned()`?

tulir: Yeah, copying from the QR code flow is probably best if there's already an equivalent section there

SpiritCroc: Picked it over, self-verifying still works. Not sure if I have a good way available to test the other path right now.

tulir (approved):

sumnerevans:
Still in the process of reviewing this...

This looks approximately correct, but the logic looks a bit weird as it's inside of the for loop for verifying all of the MACs.

I'm going to take a crack at updating this to what I think should happen, and then we can see if that still works.

sumnerevans:
@SpiritCroc can you try the latest commit? I refactored a lot of things because I realized that there were other issues with the trust codepath.

I think it's more correct now, and is more closely aligned with the logic for the QR code verification.

SpiritCroc:
Having the device trust logic within the MAC loop did confuse me too while I was looking into it - makes sense to me that that was off too, thanks for checking! I now re-tested following cases between two devices with the latest changes from the PR:

- QR verification
- Emoji verification, new device clicks the "matches" button last
- Emoji verification, new device clicks the "matches" button last

All these cases work now for me :+1:

tulir (approved):

PR #334, opened 2024-12-19
filter: make sub-structs properly nullable so omitempty works

sumnerevans:

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -19,24 +19,24 @@ const (
 // Filter is used by clients to specify how the server should filter responses to e.g. sync requests
 // Specified by: https://spec.matrix.org/v1.2/client-server-api/#filtering
 type Filter struct {
-	AccountData FilterPart  `json:"account_data,omitempty"`
+	AccountData *FilterPart `json:"account_data,omitempty"`
 	EventFields []string    `json:"event_fields,omitempty"`
 	EventFormat EventFormat `json:"event_format,omitempty"`
-	Presence    FilterPart  `json:"presence,omitempty"`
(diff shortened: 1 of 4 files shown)


tulir (approved):
Hmm, would be nicer if they weren't pointers, but I guess it's better to have omitempty working

filter.go
```diff
@@ -19,39 +19,39 @@ const (
...
 // FilterPart is used to define filtering rules for specific categories of events
 type FilterPart struct {
-	NotRooms    []id.RoomID  `json:"not_rooms,omitempty"`
-	Rooms       []id.RoomID  `json:"rooms,omitempty"`
-	Limit       int          `json:"limit,omitempty"`
-	NotSenders  []id.UserID  `json:"not_senders,omitempty"`
-	NotTypes    []event.Type `json:"not_types,omitempty"`
-	Senders     []id.UserID  `json:"senders,omitempty"`
-	Types       []event.Type `json:"types,omitempty"`
-	ContainsURL *bool        `json:"contains_url,omitempty"`
```
tulir, line 51: This has to be a pointer to allow setting it to false

sumnerevans: Do all of the booleans have to be allowed to be set to `false`?

sumnerevans: I guess the other ones default to `false`.

tulir: Yeah contains_url is the only one that is omit/true/false. For the rest, omit=false

> If true, includes only events with a url key in their content. If false, excludes those events. If omitted, url key is not considered for filtering.

PR #336, opened 2024-12-22
Support setting status_msg in Client.SetPresence

nexy7574:
Self explanatory. Needed for https://github.com/tulir/gomuks/pull/562.

### DIFF
--- a/client.go
+++ b/client.go
@@ -693,7 +693,7 @@ type ReqSync struct {
 	Since           string
 	FilterID        string
 	FullState       bool
-	SetPresence     ReqPresence
+	SetPresence     event.Presence
 	StreamResponse  bool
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -693,7 +693,7 @@ type ReqSync struct {
 	Since           string
 	FilterID        string
 	FullState       bool
-	SetPresence     event.Presence
+	SetPresence     ReqPresence
```
tulir, line 696: This change seems pointless

tulir (changes requested):

client.go
nexy7574: oh yeah 🥴
I thought it was related to setpresence for some reason

tulir (approved):

PR #339, opened 2025-01-06
bridgev2/portal: add blocking queue config

Fizzadar:
When enabled portal events always block until sent to the queue and will never be dropped.

### DIFF
--- a/bridgev2/portal.go
+++ b/bridgev2/portal.go
@@ -84,7 +84,7 @@ type Portal struct {
 	events chan portalEvent
 }
 
-const PortalEventBuffer = 64
+var PortalEventBuffer = 64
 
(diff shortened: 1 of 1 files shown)


bridgev2/portal.go
```diff
@@ -272,12 +272,16 @@ func (br *Bridge) GetExistingPortalByKey(ctx context.Context, key networkid.Port
 }
 
 func (portal *Portal) queueEvent(ctx context.Context, evt portalEvent) {
-	select {
-	case portal.events <- evt:
-	default:
-		zerolog.Ctx(ctx).Error().
-			Str("portal_id", string(portal.ID)).
-			Msg("Portal event channel is full")
+	if PortalEventBuffer == 0 {
```
tulir, line 275: PortalEventBuffer probably needs to be turned into a `var` rather than `const`

Fizzadar: Oh good spot, fixed!

PR #340, opened 2025-01-08
client: add optional media HTTP client

Fizzadar:


client.go
```diff
@@ -88,6 +88,7 @@ type Client struct {
 	Crypto        CryptoHelper
 	Verification  VerificationHelper
 	SpecVersions  *RespVersions
+	MediaClient   *http.Client // The HTTP client used for external (not matrix) media HTTP requests.
```
tulir, line 91: Probably better to call it `ExternalClient` since it's only used for external media requests, not all media requests

PR #342, opened 2025-01-09
Prevent checkpoints for double puppeted events

bradtgmurray:


bridgev2/database/upgrades/19-add-double-puppeted-to-message.sql
```diff
@@ -0,0 +1,2 @@
+-- v19 (compatible with v9+): Add double puppeted state to messages
+ALTER TABLE messages ADD COLUMN is_double_puppeted BOOLEAN NOT NULL;
```
tulir, line 2: Should leave it nullable or add a default, otherwise it's not actually compatible with v9+

bradtgmurray: Good call, making it nullable so that we can debug old rows as "not having the information", I'm assuming the database layer will convert nulls to falses in the Go struct.

tulir: It won't convert anything, you need to scan into a pointer or an `sql.NullBool` (see how the other nullable columns are scanned)

bradtgmurray: Fixed!

PR #352, opened 2025-02-12
Only accept secret shares from verified devices

bradtgmurray:


crypto/sharing.go
```diff
@@ -173,6 +173,22 @@ func (mach *OlmMachine) receiveSecret(ctx context.Context, evt *DecryptedOlmEven
 		return
 	}
 
+	// https://spec.matrix.org/v1.10/client-server-api/#msecretsend
+	// "The recipient must ensure... that the device is a verified device owned by the recipient"
+	if senderDevice, err := mach.GetOrFetchDevice(ctx, evt.Sender, evt.SenderDevice); err != nil {
```
bradtgmurray, line 178: Hrm, should I GetOrFetchDevice here or just look in the local store?

The key backup signature checking only looks in the local store
https://github.com/mautrix/go/blob/100d945d3977a3261e01090e245c9ffa3bd47af2/crypto/keybackup.go#L84

crypto/sharing.go
```diff
@@ -173,6 +173,19 @@ func (mach *OlmMachine) receiveSecret(ctx context.Context, evt *DecryptedOlmEven
...
+		log.Err(err).Msg("Failed to get or fetch sender device, rejecting secret")
+		return
+	} else if senderDevice == nil {
+		log.Warn().Msg("Unknown sender device, rejecting secret")
+		return
+	} else if !mach.IsDeviceTrusted(senderDevice) {
```
tulir, line 184: IsDeviceTrusted is actually wrong since it doesn't take a context, I should probably deprecate it or add the parameter 🤔

tulir: Added context parameter

crypto/sharing.go
tulir: It should always have the list cached, so shouldn't make a difference either way. If your own device list is somehow missing, it's probably only good that /sync is blocked until the list is fetched

sumnerevans (approved):
Looks correct to me

PR #360, opened 2025-03-09
Extend HTTP timeouts for getting the full key backup

bradtgmurray:
The default is 30 seconds, and this is way too short for users with large key backups.


tulir:
The default is 180 seconds though 🤔 https://github.com/mautrix/go/blob/main/client.go#L2525

Creating new http clients like that is bad since the connection pool is stored in the transport

bradtgmurray:
> The default is 180 seconds though 🤔 https://github.com/mautrix/go/blob/main/client.go#L2525

Hrm, maybe we just have to increase the header timeout as well? A little funky to do it across the board though...

PR #364, opened 2025-03-26
event: add Beeper transcription event definitions

SpiritCroc:


event/beeper.go
```diff
@@ -57,6 +57,24 @@ type BeeperMessageStatusEventContent struct {
...
+type BeeperRelationType string
+
+const (
+	BeeperRelTranscription BeeperRelationType = "com.beeper.transcription"
+)
+
+type BeeperRelatesTo struct {
+	EventID id.EventID         `json:"event_id,omitempty"`
+	RoomID  id.RoomID          `json:"room_id,omitempty"`
+	Type    BeeperRelationType `json:"rel_type,omitempty"`
```
sumnerevans, line 69: This should probably just be a `RelationType` and we should add `RelBeeperTranscription` as a constant.

https://github.com/mautrix/go/blob/06f200da0d1099ec9b59c50ceb754380d8a16674/event/relations.go#L17-L22

SpiritCroc: I considered that, however the official spec doesn't have a `room_id` so I went with the other approach. I'm happy to switch though - Should we prefix the `room_id` to `com.beeper.room_id` in that case?

tulir: Sumner meant the RelationType constants, not the struct

I don't remember where we landed on the struct, but adding a prefixed field and using `m.relates_to` would be fine. If we *don't* reuse it, then there will need to be extra logic for preventing encryption of `com.beeper.relates_to`. When using `m.relates_to`, it just needs a few extra getter functions and the encryption logic will automatically copy it to the unencrypted content

sumnerevans: I think we decided that since the spec says that the event must be in the same room, we can't use `m.relates_to`.

SpiritCroc: Makes sense, thanks! Pushed the according change.

I figured with the event definitions in mautrix-go, it makes sense to ensure `com.beeper.relates_to` is included in plain from mautrix-go as well rather than the SDK (what I did before), so I replicated some `RelatesTo` code for that and added it to this PR as well. Let me know if this is the correct thing to do!

tulir: Hmm, if we want mautrix-go to copy it to the unencrypted content, we should just use `m.relates_to`

SpiritCroc: So you think we should teach `m.relates_to` about `room_id` then?

tulir: Yes (or `com.beeper.room_id` to be specific)

SpiritCroc: ok, will do that then :+1:

SpiritCroc: Unfortunately the server gives me a 400 "Relations must be in the same room" then... so I guess better to stick to `com.beeper.relates_to`, probably not worth modifying the backend for that?

tulir: Hmm, are you testing with a synapse room or the hungry one? Hungry shouldn't have any validation like that. It's fine to use the other relates_to key too, then the unencrypted copying just has to be done in the SDK

SpiritCroc: Ah, I was testing with a random synapse room indeed, since that made it easier for me to check the actual encprypted vs plain event source. With the correct room it does appear to send fine, right. However my changes don't work immediately as expected so I'd probably need to grasp a hold on that room for debugging somehow, if we stick to using `m.relates_to` as is.

For your latest suggestion, would that basically be what I'm doing in the current state of the PR (in https://github.com/mautrix/go/pull/364/commits/2519eb8f11e9803250c3913364acc0e83aba1e31 ), or something different?

tulir: If using `com.beeper.relates_to`, the PR is otherwise fine, but shouldn't have the crypto stuff (so changes to `encryptmegolm.go` and `encryption.go` should be reverted and the `OptionalGetBeeperRelatesTo` interface is unnecessary)

tulir (approved):

event/beeper.go
SpiritCroc: Pushed and tested, thanks for reviewing!

sumnerevans (approved):

PR #365, opened 2025-04-02
bridgev2/legacymigrate: add post-migrate hook

sumnerevans:


bridgev2/matrix/mxmain/legacymigrate.go
```diff
@@ -239,6 +239,12 @@ func (br *BridgeMain) PostMigrate(ctx context.Context) error {
 		if err != nil {
 			zerolog.Ctx(ctx).Warn().Err(err).Stringer("room_id", portal.MXID).Msg("Failed to set service members")
 		}
+		if br.PostMigratePortal != nil {
+			err = br.PostMigratePortal(ctx, portal)
```
tulir, line 243: Probably needs to happen before the state event so bot power levels are correct. Might also need to disable the default postMigrateDMPortal call when this is set?

PR #368, opened 2025-04-09
bridgev2: make Bridge.Start() take a context

adamvy:


bridgev2/bridge.go
```diff
@@ -107,8 +107,8 @@ func (e DBUpgradeError) Unwrap() error {
 	return e.Err
 }
 
-func (br *Bridge) Start() error {
-	ctx := br.Log.WithContext(context.Background())
+func (br *Bridge) Start(ctx context.Context) error {
+	ctx = br.Log.WithContext(ctx)
```
tulir, line 111: The log could be added by the caller too, but I guess either way works

PR #369, opened 2025-04-14
Require cross signing keys to be present to verify another device

bradtgmurray:
Saw a case today where an unverified client was accepting a verification request when it didn't have any cross signing keys or wasn't even verified itself. We should just drop these like any other kind of verification request we shouldn't handle.

### DIFF
--- a/crypto/verificationhelper/verificationhelper.go
+++ b/crypto/verificationhelper/verificationhelper.go
@@ -607,6 +607,11 @@ func (vh *VerificationHelper) onVerificationRequest(ctx context.Context, evt *ev
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


bradtgmurray:
Oh no, this breaks the tests

bradtgmurray:
> This should work fine... Only one side needs to have the keys, and it could be the side that is requesting the verification.

Ah yes, I see the tests now.

I'll have to dig in deeper to figure out how we have cases of QR codes being scanned and only after the scan happens we get the `the master key is not trusted by this device, cannot verify device that does not trust the master key` error. Ideally we'd be able to figure out that the verification isn't going to work before we get to that step.

PR #371, opened 2025-04-17
Backport msc4190 from bridgev2

onestacked:
I've attempted to backport the [msc4190 support](https://github.com/mautrix/go/pull/288) form bridgev2.

The main reason for this is to allow mautrix-discord to use msc4190 before the bridgev2 rewrite.

~~I think this should work, but I'm unsure how to test this.~~

~~From what I undestand~~ The (encryption) config upgrade needs to be done in the bridge itself instead of here.
Draft [PR to mautrix-discord](https://github.com/mautrix/discord/pull/181)

### DIFF
--- a/appservice/registration.go
+++ b/appservice/registration.go
@@ -29,6 +29,7 @@ type Registration struct {
 
 	SoruEphemeralEvents bool `yaml:"de.sorunome.msc2409.push_ephemeral,omitempty" json:"de.sorunome.msc2409.push_ephemeral,omitempty"`
 	EphemeralEvents     bool `yaml:"push_ephemeral,omitempty" json:"push_ephemeral,omitempty"`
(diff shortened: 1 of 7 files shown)


tulir:
Discord is on v0.16.x, not v0.16.0.x

Easiest way to test local changes is adding `replace maunium.net/go/mautrix => ../mautrix-go` in the bridge's go.mod (assuming mautrix-go is cloned at that path)

onestacked:
> Discord is on v0.16.x, not v0.16.0.x
Changed to correct target branch

> Easiest way to test local changes is adding `replace maunium.net/go/mautrix => ../mautrix-go` in the bridge's go.mod (assuming mautrix-go is cloned at that path)
I wanted to avoid having to use the whole discord bridge, but It seems to work in my test setup with synapse & MAS.

onestacked:
@tulir can this (and the mautrix-discord change) be included in the May releases?

PR #372, opened 2025-04-23
Log invalid verification txn state (+bonus exit early on external upload cancel)

Fizzadar:


tulir (approved):
(make sure to push to main manually rather than squash merging via github)

PR #374, opened 2025-04-25
bridgev2/userlogin: delete disappearing messages when deleting portals

Fizzadar:


tulir:
It should probably just be a foreign key rather than deleting manually

bridgev2/database/upgrades/21-disappearing-message-fkey.sqlite.sql
```diff
@@ -0,0 +1,30 @@
+-- v21: Add foreign key constraint from disappearing_message.mx_room to portals.mxid
+-- transaction: off
+PRAGMA foreign_keys = OFF;
```
tulir, line 3: I don't think this migration requires disabling foreign keys or transactions

Also, portal will need a unique key on mxid (I think one can be added in sqlite without recreating the table by creating a unique index)

bridgev2/database/upgrades/00-latest.sql
```diff
@@ -128,7 +128,11 @@ CREATE TABLE disappearing_message (
 	timer        BIGINT NOT NULL,
 	disappear_at BIGINT,
 
-	PRIMARY KEY (bridge_id, mxid)
+	PRIMARY KEY (bridge_id, mxid),
+    CONSTRAINT fk_disappearing_message_portal
```
tulir, line 132: Name should be `disappearing_message_portal_fkey` to match all the existing foreign keys

bridgev2/database/upgrades/21-disappearing-message-fkey.sqlite.sql
Fizzadar: Done and done!

Fizzadar:
@tulir eventually got the right thing pushed up

bridgev2/database/upgrades/21-disappearing-message-fkey.postgres.sql
```diff
@@ -0,0 +1,2 @@
+-- v21: Add unique index on portal (bridge_id, mxid)
+CREATE UNIQUE INDEX portal_bridge_mxid_idx ON portal (bridge_id, mxid);
```
tulir, line 2: It should be doable with a single new version rather than 2, although I guess either way is fine

If it's 2 versions, this shouldn't have `.postgres.` in the file name and the versions of the second one should be fixed. New migrations also need the `compatible with` tag like the latest revision

bridgev2/database/upgrades/21-disappearing-message-fkey.sqlite.sql
```diff
@@ -0,0 +1,24 @@
...
+    disappear_at BIGINT,
+
+    PRIMARY KEY (bridge_id, mxid),
+    CONSTRAINT disappearing_message_portal_fkey
+        FOREIGN KEY (bridge_id, mx_room)
+        REFERENCES portal (bridge_id, mxid)
+        ON DELETE CASCADE
+);
+
+WITH portal_mxids AS (SELECT mxid FROM portal)
```
tulir, line 18: not sure if this is necessary, but might be

```suggestion
WITH portal_mxids AS (SELECT mxid FROM portal WHERE mxid IS NOT NULL)
```

bridgev2/database/upgrades/21-disappearing-message-fkey.postgres.sql
```diff
@@ -0,0 +1,7 @@
+-- v21: Add foreign key constraint from disappearing_message.mx_room to portals.mxid
+CREATE UNIQUE INDEX portal_bridge_mxid_idx ON portal (bridge_id, mxid);
+ALTER TABLE disappearing_message
```
tulir, line 3: Postgres also needs to delete the extra rows if there are any

```suggestion
DELETE FROM disappearing_message WHERE mx_room NOT IN (SELECT mxid FROM portal WHERE mxid IS NOT NULL);
ALTER TABLE disappearing_message
```

bridgev2/database/upgrades/21-disappearing-message-fkey.postgres.sql
```diff
@@ -0,0 +1,7 @@
+-- v21: Add foreign key constraint from disappearing_message.mx_room to portals.mxid
```
tulir, line 1: oh the compat flags are missing too

```suggestion
-- v21 (compatible with v9+): Add foreign key constraint from disappearing_message.mx_room to portals.mxid
```

bridgev2/database/upgrades/21-disappearing-message-fkey.sqlite.sql
```diff
@@ -0,0 +1,24 @@
+-- v21: Add foreign key constraint from disappearing_message.mx_room to portals.mxid
```
tulir, line 1:
```suggestion
-- v21 (compatible with v9+): Add foreign key constraint from disappearing_message.mx_room to portals.mxid
```

Fizzadar:
OK now I think we're there 🙃

tulir (approved):
Looks like it might work now

PR #376, opened 2025-05-07
client: don't attempt to make requests if the homeserver URL isn't set

Fizzadar:
Quick guard for where the client is created without using the `NewClient` method.


Fizzadar:
> Why are clients created that way 🤔

Ask whoever wrote the SDK 🤣

PR #377, opened 2025-05-07
bridgev2: add shared event handling context

Fizzadar:
This context is then passed into the network connectors handlers and message conversion functions which may require making network requests, which before this would not be canceled on bridge stop.

### DIFF
--- a/bridgev2/bridge.go
+++ b/bridgev2/bridge.go
@@ -54,6 +54,9 @@ type Bridge struct {
 
 	wakeupBackfillQueue chan struct{}
 	stopBackfillQueue   *exsync.Event
+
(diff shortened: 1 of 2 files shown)


bridgev2/bridge.go
```diff
@@ -54,6 +54,9 @@ type Bridge struct {
 
 	wakeupBackfillQueue chan struct{}
 	stopBackfillQueue   *exsync.Event
+
+	handleEventCtx context.Context
+	cancelEventCtx context.CancelFunc
```
tulir, line 59: Might be nicer to call it backgroundCtx in case it's needed for other things

tulir (approved):

PR #382, opened 2025-05-28
bridgev2/provisioning: include underlying error in failed to wait res…

Fizzadar:
…ponse


tulir (changes requested):
Should probably use `RespondWithError` (like all other endpoints already do) to only expose properly wrapped error messages rather than exposing all internal errors. Not sure why this doesn't already do that, might have been missed in some earlier refactor

PR #387, opened 2025-06-06
Add msc4193 for file spoilers

LeaPhant:
MSC4193: https://github.com/matrix-org/matrix-spec-proposals/pull/4193

This is currently rendered only by Cinny as far as I know, but useful for those on it.


tulir:
It's just a single boolean field, so easy enough to handle via the Raw field in event content

PR #390, opened 2025-06-19
verificationhelper: use static format strings

V02460:
Starting with Go version 1.24 format strings must be static. This PR fixes the one occurrence preventing mautrix from running under the new Go version.


tulir:
> preventing mautrix from running under the new Go version.

(it's a linter rule, not a compile error, it won't prevent running)

PR #399, opened 2025-08-16
event: implement disappearing timer types

KishanBagaria:
- introduces DisappearingTimerCapability, StateBeeperDisappearingTimer, BeeperDisappearingTimer
- moves DisappearingType from database to event
- UpdateDisappearingSetting: sends new StateBeeperDisappearingTimer event in room

more to come in other PRs


bridgev2/portal.go
```diff
@@ -4046,6 +4046,19 @@ func (portal *Portal) UpdateDisappearingSetting(ctx context.Context, setting dat
 	if portal.MXID == "" {
 		return true
 	}
+
+	timerType := event.DisappearingTypeAfterSend
+	if setting.Type == database.DisappearingTypeAfterRead {
+		timerType = event.DisappearingTypeAfterRead
+	} else if setting.Type == database.DisappearingTypeNone {
+		timerType = event.DisappearingTypeNone
+	}
```
tulir, line 4055: It's the same type, you don't need to map anything

KishanBagaria: good catch

tulir (approved):
Made some changes so this looks good now, I haven't tested the migration yet though (it should automatically send the state event to all rooms when a bridge is upgraded to have the capability)

PR #406, opened 2025-08-29
bridgev2/matrixinterface: add GetEvent interface for linkedin reply

rnons:


appservice/intent.go
```diff
@@ -542,3 +542,7 @@ func (intent *IntentAPI) EnsureInvited(ctx context.Context, roomID id.RoomID, us
 	}
 	return nil
 }
+
+func (intent *IntentAPI) GetEvent(ctx context.Context, roomID id.RoomID, eventID id.EventID) (*event.Event, error) {
+	return intent.Client.GetEvent(ctx, roomID, eventID)
+}
```
tulir, line 548: This is unnecessary, functions only need to be wrapped if there's additional special behavior (you can already call any Client functions on IntentAPI because it's an unnamed/embedded struct)

bridgev2/matrix/intent.go
```diff
@@ -674,3 +674,7 @@ func (as *ASIntent) MuteRoom(ctx context.Context, roomID id.RoomID, until time.T
 		})
 	}
 }
+
+func (as *ASIntent) GetEvent(ctx context.Context, roomID id.RoomID, eventID id.EventID) (*event.Event, error) {
+	return as.Matrix.GetEvent(ctx, roomID, eventID)
```
tulir, line 679: This is missing decryption and the content also need to be parsed

1. After getting event, if there's no error, call `evt.Content.ParseRaw(evt.Type)`. The error from that can be logged, but no need to return it
2. If `evt.Type` is `event.EventEncrypted`, you need to decrypt it. However, if `Bridge.Config.Encryption.DeleteKeys.RatchetOnDecrypt` is true, don't even try decrypting, just return an error. After successfully decrypting, call ParseRaw again on the result before returning it

bridgev2/matrix/intent.go
```diff
@@ -674,3 +674,23 @@ func (as *ASIntent) MuteRoom(ctx context.Context, roomID id.RoomID, until time.T
...
 }
+
+func (as *ASIntent) GetEvent(ctx context.Context, roomID id.RoomID, eventID id.EventID) (*event.Event, error) {
+	evt, err := as.Matrix.Client.GetEvent(ctx, roomID, eventID)
+	if err != nil {
+		return nil, err
+	}
+	err = evt.Content.ParseRaw(evt.Type)
+	if err != nil {
+		zerolog.Ctx(ctx).Err(err).Str("roomID", roomID.String()).Str("eventID", string(eventID)).Msg("failed to parse event content")
```
tulir, line 685: `Stringer(x)` is preferred over `Str(x.String())` (I just noticed the lint rule wasn't enabled, so enabled it on main just now)

bridgev2/matrix/intent.go
```diff
@@ -674,3 +674,23 @@ func (as *ASIntent) MuteRoom(ctx context.Context, roomID id.RoomID, until time.T
...
 }
+
+func (as *ASIntent) GetEvent(ctx context.Context, roomID id.RoomID, eventID id.EventID) (*event.Event, error) {
+	evt, err := as.Matrix.Client.GetEvent(ctx, roomID, eventID)
+	if err != nil {
+		return nil, err
+	}
+	err = evt.Content.ParseRaw(evt.Type)
+	if err != nil {
+		zerolog.Ctx(ctx).Err(err).Stringer("roomID", roomID).Stringer("eventID", eventID).Msg("failed to parse event content")
```
tulir, line 685: oops, missed that the capitalization for the log variables was wrong too (logs are json, which means snake_case everywhere)

```suggestion
		zerolog.Ctx(ctx).Err(err).Stringer("room_id", roomID).Stringer("event_id", eventID).Msg("failed to parse event content")
```

tulir (approved):
Looks fine after log variable fix

PR #417, opened 2025-09-24
expand syncParticipants to ensure base powerlevels, add ChatInfoFetchingNetworkAPI

adamvy:


bridgev2/portal.go
```diff
@@ -4046,7 +4046,16 @@ func (portal *Portal) syncParticipants(
 		return fmt.Errorf("failed to get current members: %w", err)
 	}
 	delete(currentMembers, portal.Bridge.Bot.GetMXID())
-	powerChanged := members.PowerLevels.Apply(portal.Bridge.Bot.GetMXID(), currentPower)
+	// Ensure baseline power levels match initial room creation defaults.
+	// See createMatrixRoomInLoop: we enforce bot at 9001 and lock down
```
adamvy, line 4050: Is this right/safe?

tulir: It's unsafe for rooms that aren't created by the bridge and doesn't seem necessary for local bridges either

PR #438, opened 2025-12-05
client: add missing retry cancel check while backing off requests

Fizzadar:


client.go
```diff
@@ -614,7 +614,9 @@ func (cli *Client) doRetry(
 	select {
 	case <-time.After(backoff):
 	case <-req.Context().Done():
-		return nil, nil, req.Context().Err()
+		if !errors.Is(context.Cause(req.Context()), ErrContextCancelRetry) {
+			return nil, nil, req.Context().Err()
+		}
```
tulir: Hmm, this will short circuit the backoff, but maybe that's fine

PR #454, opened 2026-01-20
event: add action message content for rich call notifications

SpiritCroc:

### DIFF
--- a/event/beeper.go
+++ b/event/beeper.go
@@ -175,8 +175,8 @@ const (
 type BeeperActionMessageCallType string
 
 const (
-	BeeperActionMessageCallTypeVoice BeeperActionMessageType = "voice"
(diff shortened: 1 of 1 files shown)


event/beeper.go
```diff
@@ -166,6 +166,11 @@ type BeeperPerMessageProfile struct {
 	HasFallback bool                 `json:"has_fallback,omitempty"`
 }
 
+type BeeperActionMessage struct {
+	Type     string `json:"type"`
+	CallType string `json:"call_type,omitempty"`
```
tulir: custom string types with constant values are preferred for things like these

```go
type BeeperActionMessageType string

const (
	BeeperActionMessageCall BeeperActionMessageType = "call"
)
```

PR #465, opened 2026-02-23
bridgev2/login: add attachments option to user input step type

radon-at-beeper:
Experimental. Goes with https://github.com/mautrix/meta/pull/226.

Something like this:

image

### DIFF
--- a/bridgev2/commands/login.go
+++ b/bridgev2/commands/login.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/skip2/go-qrcode"
 	"go.mau.fi/util/curl"
-	"go.mau.fi/util/exmime"
 
 	"maunium.net/go/mautrix/bridgev2"
 	"maunium.net/go/mautrix/bridgev2/networkid"


bridgev2/login.go
```diff
@@ -271,6 +279,16 @@ func (f *LoginInputDataField) FillDefaultValidate() {
...
+
+	// Attachments to display alongside the input fields.
+	Attachments []*LoginUserInputAttachment `json:"attachments"`
+}
+
+type LoginUserInputAttachment struct {
+	Type     event.MessageType
+	Filename string
+	MimeType string
+	Content  []byte
```
tulir: These should have json struct tags and all this should be added to the provisioning API spec in `bridgev2/matrix/provisioning.yaml`

bridgev2/login.go
```diff
@@ -58,6 +59,11 @@ type LoginProcessCookies interface {
 	SubmitCookies(ctx context.Context, cookies map[string]string) (*LoginStep, error)
 }
 
+type LoginProcessCaptcha interface {
+	LoginProcess
+	SubmitCaptcha(ctx context.Context, code string) (*LoginStep, error)
+}
```
tulir: This is unused now

bridgev2/commands/login.go
```diff
@@ -273,6 +274,34 @@ func sendQR(ce *Event, qr string, prevEventID *id.EventID) error {
 	return nil
 }
 
+func sendUserInputAttachments(ce *Event, atts []*bridgev2.LoginUserInputAttachment) error {
+	for _, att := range atts {
+		filename := "captcha" + exmime.ExtensionFromMimetype(att.Info.MimeType)
```
tulir: oh, this isn't using the filename in the struct even though one was added there

PR #487, opened 2026-04-14
client: support MSC4446 for moving `m.fully_read` backwards

SpiritCroc:
https://github.com/matrix-org/matrix-spec-proposals/pull/4446

### Checklist


versions_test.go
tulir, line 1: This generally shouldn't be modified at all

tulir (approved):

PR #488, opened 2026-04-14
bridgev2/matrix: Propagate start errors to caller

hifi:
Avoid calling os.Exit in the Matrix connector and instead pass it down as an error for mxmain to do.

There are still some places where os.Exit is being called but this covers most of them.

Also fixes the reconnection loop in ensureConnection which would not stop if the start context is canceled.

### DIFF
--- a/bridgev2/matrix/connector.go
+++ b/bridgev2/matrix/connector.go
@@ -14,7 +14,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"os"
 	"regexp"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


bridgev2/matrix/connector.go
```diff
@@ -365,15 +366,22 @@ func (br *Connector) ensureConnection(ctx context.Context) {
 				err = br.Bot.EnsureRegistered(ctx)
 				if err != nil {
 					br.logInitialRequestError(err, "Failed to register after /versions failed with M_FORBIDDEN")
-					os.Exit(16)
+					return errors.Join(err, ExitError{16})
```
tulir: `errors.Join` isn't nice, `fmt.Errorf("%w: failed to register after /versions failed with M_FORBIDDEN: %w", ExitError{16}, err)` is better if the error is ever printed

(also applies to all the other join returns)

hifi: Done.

bridgev2/matrix/connector.go
```diff
@@ -395,38 +403,37 @@ func (br *Connector) ensureConnection(ctx context.Context) {
...
 	}
 
 	if br.Websocket {
 		br.Log.Debug().Msg("Websocket mode: no need to check status of homeserver -> bridge connection")
-		return
 	} else if !br.SpecVersions.Supports(mautrix.FeatureAppservicePing) {
 		br.Log.Debug().Msg("Homeserver does not support checking status of homeserver -> bridge connection")
-		return
+	} else {
+		br.Bot.EnsureAppserviceConnection(ctx)
```
tulir: This can currently exit too, but I think it can be replaced with a boolean return since there's only one exit code, then `if !br.Bot.EnsureAppserviceConnection(ctx) { return ExitError{13} }`

hifi: Changed.

tulir (approved):

PR #492, opened 2026-04-22
client: Rework network request retries

adamvy:
Add a simple trigger using exsync.Event to abort and retry network requests.

Before this logic was somewhat split between beeper internal code and mautrix/go.  This resulting a lot of redundant book-keeping on the beeper side to interact nicely with mautrix.

New logic here moves the per-request context creation into mautrix directly hooked into the retry logic.  The new logic only applies if a RequestRetryTrigger is provided.

Also included are a bunch of ai generated tests, but happy to remove or simplify them if you want.

Beeper side of this is https://github.com/beeper/beeper/pull/2466


client.go
```diff
@@ -738,6 +744,46 @@ func ParseErrorResponse(req *http.Request, res *http.Response) ([]byte, error) {
...
+
+	attemptCtx, cancel := context.WithCancelCause(req.Context())
+	var finishOnce sync.Once
+	finish := func() {
+		finishOnce.Do(func() {
+			cancel(context.Canceled)
+		})
+	}
+
+	resetChan := cli.RequestRetryTrigger.GetChan()
```
adamvy: I suppose I could just do RequestRetryTrigger.Wait(attemptCtx) in the goroutine?

client.go
```diff
@@ -738,6 +744,42 @@ func ParseErrorResponse(req *http.Request, res *http.Response) ([]byte, error) {
...
+		if cli.RequestRetryTrigger.Wait(attemptCtx) == nil {
+			cancel(ErrContextCancelRetry)
+		}
+	}()
+
+	return req.WithContext(attemptCtx), func() {
+		cleanupOnce.Do(func() {
+			cancel(context.Canceled)
+		})
+	}
```
tulir:
```suggestion
	return req.WithContext(attemptCtx), sync.OnceFunc(func() {
		cancel(context.Canceled)
	})
```

client.go
```diff
@@ -738,6 +744,42 @@ func ParseErrorResponse(req *http.Request, res *http.Response) ([]byte, error) {
...
 }
 
+func (cli *Client) prepareRequestAttempt(req *http.Request) (*http.Request, func()) {
+	// if there's no retry trigger, nothing to do
+	if cli.RequestRetryTrigger == nil {
+		return req, func() {}
+	}
+
+	attemptCtx, cancel := context.WithCancelCause(req.Context())
+	var cleanupOnce sync.Once
```
tulir:
```suggestion
```

client.go
```diff
@@ -96,6 +98,10 @@ type Client struct {
 	RequestHook  func(req *http.Request)
 	ResponseHook func(req *http.Request, resp *http.Response, err error, duration time.Duration)
 
+	RequestRetryTrigger *exsync.Event
+
+	// Deprecated: this hook is no longer used by mautrix internals and is kept only
```
adamvy: I suppose we could just delete this and consumers can either upgrade to the new behaviour, or just drop the functionality entirely.

Is anyone outside beeper using this atm?

client.go
```diff
@@ -748,30 +786,46 @@ func (cli *Client) executeCompiledRequest(
...
+	res, err := client.Do(attemptReq)
 	duration := time.Since(startTime)
-	if res != nil && !dontReadResponse {
-		defer res.Body.Close()
+	if res != nil {
+		// Cleanup the child attempt context once the body is closed
+		res.Body = cleanupReadCloser{
+			ReadCloser: res.Body,
+			cleanup:    cleanup,
+		}
```
tulir: Wrapping readers/writers isn't very nice since it masks anything not included in the interface, would be nicer to only do it when `cleanup` is needed.

Should also check if http response bodies usually implement `WriterTo` and include that in the proxy if they do (see how `io.NopCloser` works)

adamvy:
I suppose an alternative would be to just do this all at the sdk layer and not involve mautrix at all?  LIke a whole other retry wrapper in sdk, unsure how I feel about that yet.

client.go
adamvy: done

tulir (approved):

PR #493, opened 2026-04-24
bridgev2/portal: persist portal row before CreateRoom when using deterministic room IDs

iFixRobots:
## Problem

When a bridge's `Bot.CreateRoom` implementation writes state events to a local store (e.g. an on-device bridge with a SQLite-backed send queue), a process interruption between `CreateRoom` and `portal.Save` leaves state events persisted with no matching portal row. With no portal row, per-bridge recovery code cannot find the portal to re-emit or fix state for, so the room stays half-broken until a manual intervention (typically a full bridge re-login).

This is the failure shape we've been tracking at Beeper across every local bridge network — users report chats that can't be archived, muted, tagged, or opened because the state events that `EnsureLocalRoomExists` and pagination require are missing from `local_current_state_events` while the portal is still indexed upstream.

## Fix

Re-order the critical section in `createMatrixRoomInLoop` so that when the caller has supplied a deterministic `BeeperLocalRoomID`, the portal row is persisted *before* `CreateRoom` is invoked. If `CreateRoom` is interrupted after that point, the failure shape is inverted: the portal row exists with no state events, which per-bridge recovery code can detect and heal by re-emitting state events against the known portal.

Preserves the existing path for two cases:
- `BeeperLocalRoomID` is empty → server assigns the room ID, so we can only save after `CreateRoom` returns. Original ordering kept.
- `BeeperLocalRoomID` is set but the server ignores it and assigns a different ID → the portal row is corrected after `CreateRoom` returns.

No change in happy-path behaviour for deployments where the server honours `BeeperLocalRoomID` — the same three DB writes happen in a different order.

## Complements a bridge-layer recovery fix

The recovery side of this lives in the client SDK, where `RepairPortalState` re-emits `m.room.create` and `m.bridge` via `LocalMatrixIntent.CreateRoom` when it detects that those rows are missing for a portal the framework still knows about. With this reorder landing in mautrix-go, the recovery code's failure-shape assumption (portal exists in framework, state events missing locally) becomes the only shape we can end up in, instead of the harder-to-recover reverse.

### DIFF
--- a/bridgev2/portal.go
+++ b/bridgev2/portal.go
@@ -5267,25 +5267,81 @@ func (portal *Portal) createMatrixRoomInLoop(ctx context.Context, source *UserLo
 	if cancellableCtx.Err() != nil {
 		return cancellableCtx.Err()
 	}
+
+	// If the caller supplied a deterministic BeeperLocalRoomID, persist the
+	// portal row before sending the create request. Bridges that write state
+	// events to a local store (e.g. on-device local bridges backed by SQLite
+	// via a send queue) can otherwise end up with state events persisted but
+	// no portal row if the process is interrupted between CreateRoom and
+	// Save. Pre-persisting flips that failure mode: a partial write leaves a
+	// portal row without state events, which per-bridge recovery code can
+	// detect and heal by re-emitting state events against the known portal.
+	//
+	// When BeeperLocalRoomID is empty the server assigns the room ID, so we
+	// can only save after CreateRoom returns (original path, preserved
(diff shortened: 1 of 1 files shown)


tulir:
Not entirely sure if this is ever needed. If the app crashes in the middle of portal creation, it means the event that triggered the portal creation in the first place also failed. If it was a message, it should already retry itself on the next run. If it's post-login sync, then the entire sync probably has to restart

iFixRobots:
Re-reading this with your comment in mind, you're right. The remaining concern I had is already covered by the reactive repair in 2506. Closing.

PR #494, opened 2026-04-24
bridgev2/portal: re-emit capability anchors on UpdateCapabilities when homeserver state is missing

iFixRobots:
## Problem

`(*Portal).UpdateCapabilities` has two short-circuits that skip re-emission without verifying the capability anchor events still exist on the homeserver:

```go
if !implicit && time.Since(portal.lastCapUpdate) < 24*time.Hour {
    return false
}
...
if capID == portal.CapState.ID {
    return false
}
```

Both fire on healthy portals (correct) and broken portals (bug). When `m.bridge`, `uk.half-shot.bridge`, or `com.beeper.room_features` goes missing from the homeserver but the capability set hasn't changed, the bridge never notices and never re-emits. Clients then fall back to partial/empty capability information and disable feature-gated UI.

I've been tracking this across Beeper's local bridges where it manifests most visibly as "missing Reply" on Desktop, but the mechanism is general — any path that loses a state event on the homeserver can hit this.

## Fix

Add `anyCapabilityAnchorMissing`, which uses the existing optional `MatrixConnectorWithArbitraryRoomState.GetStateEvent` interface to check the three anchor types. When any is missing, `UpdateCapabilities`:

- bypasses the 24h throttle
- bypasses the `capID == CapState.ID` short-circuit
- calls `UpdateBridgeInfo` to re-emit the `m.bridge` / `uk.half-shot.bridge` pair
- proceeds through the existing `sendRoomMeta(..., StateBeeperRoomFeatures, ...)` path to re-emit the capabilities event

Connectors that don't implement `MatrixConnectorWithArbitraryRoomState` fall back to the existing behavior — no self-heal, no regression. Lookup errors on individual anchors are logged and treated as "not missing" so a transient failure doesn't force a re-emit storm.

## Who this helps

Any deployment whose homeserver can drop or fail to persist state events and whose matrix connector can answer `GetStateEvent` lookups. The standard `bridgev2/matrix.Connector` already implements it, so upstream deployments get self-heal for free once this lands.

## Context

On the Beeper side this pairs with a client SDK change that implements `GetStateEvent` on the local matrix connector, so on-device local bridges benefit too. That's a separate PR in the beeper-client-sdk repo and has no ordering dependency with this one — connectors opt in by implementing the interface, and the interface already exists upstream.

### DIFF
--- a/bridgev2/portal.go
+++ b/bridgev2/portal.go
@@ -4319,22 +4319,41 @@ func (portal *Portal) UpdateBridgeInfo(ctx context.Context) {
 func (portal *Portal) UpdateCapabilities(ctx context.Context, source *UserLogin, implicit bool) bool {
 	if portal.MXID == "" {
 		return false
-	} else if !implicit && time.Since(portal.lastCapUpdate) < 24*time.Hour {
-		return false
-	} else if portal.CapState.ID != "" && source.ID != portal.CapState.Source && source.ID != portal.Receiver {
+	}
+	// Check whether capability anchor events (m.bridge, uk.half-shot.bridge,
+	// com.beeper.room_features) still exist on the homeserver. If any is
+	// missing, bypass the throttle and capID short-circuit below so the full
+	// anchor set gets re-emitted. Without this check, a portal whose state
+	// events got lost on the server would never self-heal here — the bridge
+	// would happily return early on both "nothing changed" signals while the
+	// client sees no anchors and disables feature-gated UI.
+	anchorMissing := portal.anyCapabilityAnchorMissing(ctx)
+	if !anchorMissing {
+		if !implicit && time.Since(portal.lastCapUpdate) < 24*time.Hour {
+			return false
+		}
+	}
+	if portal.CapState.ID != "" && source.ID != portal.CapState.Source && source.ID != portal.Receiver {
 		// TODO allow capability state source to change if the old user login is removed from the portal
 		return false
 	}
 	caps := source.Client.GetCapabilities(ctx, portal)
 	capID := caps.GetID()
-	if capID == portal.CapState.ID {
+	if !anchorMissing && capID == portal.CapState.ID {
 		return false
 	}
 	zerolog.Ctx(ctx).Debug().
 		Str("user_login_id", string(source.ID)).
 		Str("old_id", portal.CapState.ID).
 		Str("new_id", capID).
+		Bool("anchor_missing", anchorMissing).
 		Msg("Sending new room capability event")
+	// If the caps event itself wasn't the only thing missing, also re-emit
+	// the bridge info pair so the full anchor set on the homeserver matches
+	// what a fresh portal would have.
+	if anchorMissing {
+		portal.UpdateBridgeInfo(ctx)
+	}
 	success := portal.sendRoomMeta(ctx, nil, time.Now(), event.StateBeeperRoomFeatures, portal.getBridgeInfoStateKey(), caps, false, nil)
 	if !success {
 		return false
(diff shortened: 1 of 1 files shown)


tulir:
There's never any reason for events to go missing, so this shouldn't be needed (especially since the sdk side is just querying the local db, not even checking what the app has)

iFixRobots:
@tulir You're right this shouldn't live permanently in UpdateCapabilities. Two things to clarify before closing or narrowing:

On "sdk side is just querying the local db": for local-bridge rooms the local DB is what the client reads, right? `rooms_api.go:454` routes fetchRoomStateEvent through sdkDB.LocalEvents.GetCurrentRoomStateEvent for any shared.IsLocalBridgeRoom ID, same table LocalMatrixConnector.GetStateEvent reads in #2512. Not a self-check against a private cache. 

Your point applies to the standard mautrix Connector where the bridge's local store can diverge from /state, but that divergence doesn't exist on local-bridge.

On "there's never any reason for events to go missing": harder to rebut, I don't have a smoking gun. But specific write-path places where partial state on disk is are definitely there, take a look:

- MatrixSendQueue.loop (sendqueue.go:144-213) fires OnSync before saveLocalPayload. Kill the app between those steps, client believes state exists, DB doesn't.

- `saveLocalPayload (:399-438)` walks every state event with individual saveLocalEvent calls, no enclosing transaction, returns on first error. One Upsert failure mid-payload lands N anchors and drops the others. Portal row already saved by bridgev2 earlier, so "portal indexed, anchors partially missing" exactly as PLAT-36179 reports.

- `update_local_current_state` is AFTER INSERT ... WHEN NEW.state_key IS NOT NULL (00-latest-schema.sql:121). The upsert takes the DO UPDATE path when event_id exists, and SQLite fires AFTER UPDATE triggers, not AFTER INSERT. So if a row survives in local_events but got pruned from local_current_state_events, re-inserting the same event_id does not re-populate the index. RepairPortalState uses a deterministic event_id for m.room.create, so repair itself may silently no-op there. No reproducer yet, flagging as risk before I claim #2506 is complete.

- `LocalMatrixIntent.DeleteRoom (:618-655)` explicitly notes failures land in inconsistent state.

None of these are fixed by #494. #494 only detects the result, which is your point.

Narrower shape: keep #494 but only run the anchor check when the 24h throttle would have short-circuited. Cuts cost from 3 state lookups per call to 3 once per 24h per portal. Effectively free on both local-bridge (SQLite PK) and cloud (HTTP /state). Still catches the mid-session gap the startup audit can't (user sees Reply disappear at hour 5 of a 10-hour session, doesn't restart).

```
goif !implicit && time.Since(portal.lastCapUpdate) < 24*time.Hour {
    if !portal.anyCapabilityAnchorMissing(ctx) {
        return false
    }
}
// ... existing source/Receiver guard ...
if capID == portal.CapState.ID && !portal.anyCapabilityAnchorMissing(ctx) {
    return false
}
```

If still not worth the complexity, I'll close #494 and fold into reactive repair. But I'd rather not.

tulir:
`portal.CapState` is set last, after a successful `sendRoomMeta` call. There's no path where `CapState` gets updated without the event being saved. The next `UpdateCapabilities` call will already always retry if the previous one failed.

Basically all database stuff in bridges is architected that way: send data to homeserver first, save to bridge db after successful response. If it fails at any stage, the homeserver/sdk may or may not have actually persisted the data, but the bridge won't have flagged it as a success, so it'll retry to get into a consistent state (resending something that was already persisted is assumed to be a no-op).

> for local-bridge rooms the local DB is what the client reads, right?

The apps and SDK have separate DBs (for now). SDK is effectively the homeserver

iFixRobots:
Thanks, Tulir. Closing this. I had that conflated.

PR #512, opened 2026-06-05
bridgev2: scaffold MSC4350 config flag

benmichael:
**Status:** Draft, scaffolding only — sprint 1 of a longer implementation effort. Not yet ready for review.

## What this PR does
Adds an `encryption.msc4350` config flag (default `false`) that will gate the bridge-side implementation of [MSC4350: Permitting encryption impersonation for appservices](https://github.com/matrix-org/matrix-spec-proposals/pull/4350).

This commit only introduces the flag and its docstring in the canonical example config. No behaviour changes; with the flag off (the default), the bridge behaves exactly as today.

## Why MSC4350
Currently mautrix bridges sign all encrypted events with a single bridge bot device regardless of which ghost is the event sender. Under MSC4153 (Invisible Crypto), strict clients are starting to refuse such events, and Element already warns "The sender of the event does not match the owner of the device that sent it." MSC4350 legitimizes the pattern via an `impersonator` field on the ghost's device entry.

## Roadmap
Subsequent commits on this branch will add, in this order:

1. Extend the `id.DeviceKeys` model to serialize / deserialize the unstable `fi.mau.msc4350.impersonator` field.
2. In `bridgev2/matrix/crypto.go`, after the bridge bot device is created and cross-signed, ensure each ghost user has an impersonatable device pointing at the bot via MSC4326 device masquerading.
3. Double-puppet variant: similar but additionally signed by the user's self-signing key. Requires user interaction or recovery key — gated separately.
4. Tests.

Companion synapse work: https://github.com/element-hq/synapse/pull/19827

## Engagement
@tulir — opening this draft to anchor the work and stay aligned with your direction. If you'd prefer to implement MSC4350 yourself, or have a different architecture in mind than what's outlined in the roadmap, please let me know **before** I write more code; I'd rather adapt now than duplicate effort. Otherwise I'll keep moving and ping again when there's substantive behaviour to review.

## Related
- MSC4350 spec: matrix-org/matrix-spec-proposals#4350
- MSC4153 invisible crypto: matrix-org/matrix-spec-proposals#4153
- MSC4326 device masquerading, now stable in synapse: element-hq/synapse#19033

### DIFF
--- a/bridgev2/bridgeconfig/encryption.go
+++ b/bridgev2/bridgeconfig/encryption.go
@@ -16,8 +16,19 @@ type EncryptionConfig struct {
 	Require    bool `yaml:"require"`
 	Appservice bool `yaml:"appservice"`
 	MSC4190    bool `yaml:"msc4190"`
-	MSC4392    bool `yaml:"msc4392"`
-	SelfSign   bool `yaml:"self_sign"`
+	// MSC4350: Permitting encryption impersonation for appservices.
+	// When enabled, the bridge will create an "impersonatable" device for
+	// each ghost user (and double-puppeted real user) pointing at the
+	// bridge bot's device as the impersonator. This lets recipient clients
+	// legitimately accept events signed by the bridge bot's device when
+	// attributed to a ghost, removing the "sender doesn't match device
+	// owner" warning. Requires homeserver support for MSC4350 (synapse
+	// experimental_features.msc4350_enabled) and a cross-signed bridge bot
+	// device (encryption.self_sign).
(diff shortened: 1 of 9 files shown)


benmichael:
Closing this experimental fork-based implementation. Verified the server-side path works against my home cluster (mautrix-whatsapp bridge), but client support needs the matching MSC4350 patches across matrix-rust-sdk + matrix-js-sdk + the Element clients before the warning actually clears for end users — that's a much bigger upstream coordination effort than I can carry alone. Subscribed to https://github.com/matrix-org/matrix-spec-proposals/pull/4350 to track progress. Hopefully useful as prior art if anyone picks this up.

PR #517, opened 2026-06-16
portalbackfill: keep never-bridged messages in forward backfill dedup

Fizzadar:
See https://github.com/mautrix/meta/pull/286

But basically - aggressive dupe should probably not do prefix deleting as it breaks the "check each message" approach, which in meta's case can lead to permanently lost messages.

### DIFF
--- a/bridgev2/portalbackfill.go
+++ b/bridgev2/portalbackfill.go
@@ -249,10 +249,8 @@ func (portal *Portal) cutoffMessages(ctx context.Context, messages []*BackfillMe
 		return messages
 	}
 	if forward && !aggressiveDedup {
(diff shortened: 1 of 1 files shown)


bridgev2/portalbackfill.go
```diff
@@ -265,7 +267,7 @@ func (portal *Portal) cutoffMessages(ctx context.Context, messages []*BackfillMe
 				Msg("Cutting off forward backfill messages older than latest bridged message")
 			messages = messages[cutoff+1:]
 		}
-	} else {
+	} else if !forward {
```
Fizzadar, line 270: Should we apply the aggressiveDedup guard here too so both forward/backward are the same (aggressive is forced and any prefix shortcuts disabled).

PR #518, opened 2026-06-16
bridgev2: always include DM peer in initial member list

Fizzadar:
When a DM portal is re-created and the network's resync member map omits the remote peer (conveyed only via OtherUserID), the peer never gets an m.room.member event, leaving clients without a DM hero (missing title/avatar).


tulir:
Probably a bridge bug

Fizzadar:
> Probably a bridge bug

Yep, found it => https://github.com/mautrix/meta/pull/287

PR #519, opened 2026-06-18
bridgev2/portal: don't suppress send errors during shutdown

Fizzadar:
Without this there's a short window during bridge shutdown when any sends will be ignored with status=success.


bridgev2/portal.go
```diff
@@ -587,7 +584,8 @@ func (portal *Portal) handleSingleEvent(ctx context.Context, rawEvt any, doneCal
 		}
 		res = portal.handleMatrixEvent(ctx, evt.sender, evt.evt, isStateRequest)
 		if res.SendMSS {
-			if res.Error != nil {
+			// Suppress errors caused by/during bridge shutdown, these should be redelivered
+			if res.Error != nil && !errors.Is(res.Error, portal.backgroundCtx.Err()) {
```
tulir: It should probably not (try to) send MSS at all in this case 🤔

Fizzadar: Done!

PR #522, opened 2026-07-01
bridgev2/portal: delete rooms in connector before portal removal

Fizzadar:
This will prevent situations where the connector (Matrix) has a room that can't be used because the bridge no longer has it. This does open the door to the inverse problem, but that should generally be invisible to users.

The proper fix here is a delete queue and waiting, this is a temporary workaround.


Fizzadar:
Problem here is it flips the issue over but it's entirely possible, bridge might end up sending messages to rooms the matrix side has no idea about. A proper delete queue is really the only way forward.

PR #523, opened 2026-07-07
bridgev2/portal: use prev state bad credentials message for send failures when not logged in

Fizzadar:
This gives the user a better description of the situation vs a generic you are not logged in error.

### DIFF
--- a/bridgev2/portal.go
+++ b/bridgev2/portal.go
@@ -32,6 +32,7 @@ import (
 	"maunium.net/go/mautrix"
 	"maunium.net/go/mautrix/bridgev2/database"
 	"maunium.net/go/mautrix/bridgev2/networkid"
+	"maunium.net/go/mautrix/bridgev2/status"
 	"maunium.net/go/mautrix/event"
 	"maunium.net/go/mautrix/id"
 )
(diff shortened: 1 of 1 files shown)


Fizzadar:
Switched to using `FindPreferredLogin` passing the login, even if rejected, back to the caller.

bridgev2/portal.go
```diff
@@ -665,12 +666,12 @@ func (portal *Portal) FindPreferredLogin(ctx context.Context, user *User, allowR
...
+			return login, nil, fmt.Errorf("%w (receiver login is not logged in)", ErrNotLoggedIn)
 		} else if login.UserMXID != user.MXID {
 			if allowRelay && portal.Relay != nil {
 				return nil, nil, nil
 			}
-			return nil, nil, fmt.Errorf("%w (relay not set and receiver login is owned by %s, not %s)", ErrNotLoggedIn, login.UserMXID, user.MXID)
+			return login, nil, fmt.Errorf("%w (relay not set and receiver login is owned by %s, not %s)", ErrNotLoggedIn, login.UserMXID, user.MXID)
 		}
 		up, err := portal.Bridge.DB.UserPortal.Get(ctx, login.UserLogin, portal.PortalKey)
 		return login, up, err
```
tulir, line 677: There are other cases below, I think the `no usable logins found` at the bottom is the one that non-split group portals will hit

tulir (approved):

PR #525, opened 2026-07-14
bridgev2/login: add `LoginCookiesParams.Hidden` boolean field

Fizzadar:


tulir:
Should also add it to the OpenAPI spec in bridgev2/matrix/provisioning.yaml

tulir (approved):

PR #528, opened 2026-07-24
oauth: Add optional Scope field to TokenResponse

timedoutuk:
https://datatracker.ietf.org/doc/html/rfc6749#section-5.1 - `scope` is REQUIRED if the server returns a scope differing from the client's request

### Checklist

### DIFF
--- a/oauth/types.go
+++ b/oauth/types.go
@@ -143,6 +143,7 @@ type TokenResponse struct {
 	TokenType    string           `json:"token_type"`
 	ExpiresIn    jsontime.Seconds `json:"expires_in"`
 	RefreshToken string           `json:"refresh_token,omitempty"`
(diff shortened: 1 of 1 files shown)


oauth/types.go
```diff
@@ -143,6 +143,7 @@ type TokenResponse struct {
 	TokenType    string           `json:"token_type"`
 	ExpiresIn    jsontime.Seconds `json:"expires_in"`
 	RefreshToken string           `json:"refresh_token,omitempty"`
+	Scope        Scope            `json:"scope,omitempty"`
```
tulir: Is this a space-separated scope list rather than a single scope? Should probably use a different type for that 🤔

timedoutuk: hm good point actually. `ScopeGroup (string)`?

tulir (approved):

PR #531, opened 2026-07-27
bridgev2: add client HTTP login step

highesttt:
### Checklist

## Summary

* add a typed `client_http` login step with byte bodies and multi-value HTTP headers
* add provisioning submission, request ID and response validation, and OpenAPI definitions
* make command-based login fail gracefully when a flow requires client-side HTTP

## Tests

* `go test ./bridgev2 ./bridgev2/commands ./bridgev2/matrix`
* `openapi-spec-validator bridgev2/matrix/provisioning.yaml`
* `go test ./...` (all bridgev2 packages pass; the Windows run still hits the existing Unix-socket path and libolm describe test failures)


bridgev2/login_client_http_test.go
```diff
@@ -0,0 +1,69 @@
+package bridgev2
```
Fizzadar: I think we can drop these tests they have no value.

bridgev2/matrix/provisioninglogin.go
```diff
@@ -177,6 +186,7 @@ func (prov *ProvisioningAPI) doLoginStep(
 	expectedID string,
 	params map[string]string,
 	rawParams json.RawMessage,
+	clientHTTPResponse *bridgev2.LoginClientHTTPResponse,
```
Fizzadar: Nit: I think we should just pass the JSON through here in `rawParams` and expand to `LoginClientHTTPResponse` inside the switch

Fizzadar (changes requested):
Couple of changes, looking good overall!

Feels weird to have the response be typed out (vs cookies is just a map) but it makes sense because the response is complex.

Fizzadar (approved):
LGTM

PR #532, opened 2026-07-28
bridgev2/portal: don't fail message handling when the event is already bridged

Fizzadar:
When a connector re-IDs an existing message row (mautrix-signal does this on edits, pointing the row at the edit's timestamp so chained edit targets resolve), a redelivery of the original network message misses the duplicate check in `handleRemoteMessage`, which keys on the network ID.

With an intent that generates deterministic event IDs the resend then returns the row's existing mxid, so the insert dies on `message_mxid_unique` and the whole remote event is reported as failed — on Signal that means the envelope is never acked and the receive queue stays wedged until re-link.

This treats an insert conflicting with an existing row for the same mxid as an already-handled duplicate instead of a database error.

### DIFF
--- a/bridgev2/portal.go
+++ b/bridgev2/portal.go
@@ -2931,6 +2931,15 @@ func (portal *Portal) sendConvertedMessage(
 		}
 		err := portal.Bridge.DB.Message.Insert(ctx, dbMessage)
 		if err != nil {
+			existing, dupErr := portal.Bridge.DB.Message.GetPartByMXID(ctx, dbMessage.MXID)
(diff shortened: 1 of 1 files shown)


tulir:
Hmm, checking the mxid only works for local, so it'd probably be better if the signal bridge inserted a stub message row for each old revision

Fizzadar:
> Hmm, checking the mxid only works for local, so it'd probably be better if the signal bridge inserted a stub message row for each old revision

Yep, much better: https://github.com/mautrix/signal/pull/661

PR #534, opened 2026-07-29
bridgev2: add configurable debounce for transient connection states

Fizzadar:
At our scale these little flaps tend to generate pretty insane numbers of useless state records.

### DIFF
--- a/bridgev2/matrix/mxmain/example-config.yaml
+++ b/bridgev2/matrix/mxmain/example-config.yaml
@@ -26,11 +26,8 @@ bridge:
     # These contain the same data that can be posted to an external HTTP server using homeserver -> status_endpoint.
     # Allowed values: none, errors, all
     bridge_status_notices: errors
-    # How long should TRANSIENT_DISCONNECT and CONNECTING states be held back before being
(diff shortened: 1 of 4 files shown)


bridgev2/matrix/mxmain/example-config.yaml
```diff
@@ -26,6 +26,12 @@ bridge:
     # These contain the same data that can be posted to an external HTTP server using homeserver -> status_endpoint.
     # Allowed values: none, errors, all
     bridge_status_notices: errors
+    # How long should TRANSIENT_DISCONNECT and CONNECTING states be held back before being
+    # reported? Most disconnects recover within a couple of seconds, and debouncing means a
+    # login that drops and reconnects within the delay reports no state changes at all.
+    # Whatever state the login is in when the delay passes is reported as normal, and errors
+    # are never delayed. 0s disables the debounce.
```
tulir:
```suggestion
    # How long should TRANSIENT_DISCONNECT and CONNECTING states be held back before being reported?
    # If another state update occurs while waiting, the previous state won't be reported at all.
```

PR #535, opened 2026-07-29
bridgev2/portal: fix stale ghost displayname in member events

Fizzadar:
A room member event whose displayname has drifted from the ghost profile is never repaired: the membership value is unchanged so no member event is sent, and when one is sent it copies the stale cached displayname straight back.

Sync joined ghosts whose name differs from the room's current member state, take the displayname and avatar from the ghost when rewriting an existing member event, and skip the invite downgrade for members who are already joined.

### DIFF
--- a/bridgev2/portal.go
+++ b/bridgev2/portal.go
@@ -4706,16 +4706,20 @@ func (portal *Portal) syncParticipants(
 			raw["com.beeper.exclude_from_timeline"] = true
 		}
 	}
-	syncUser := func(extraUserID id.UserID, member ChatMember, intent MatrixAPI) bool {
+	syncUser := func(extraUserID id.UserID, member ChatMember, intent MatrixAPI, ghost *Ghost) bool {
 		if member.Membership == "" {
 			member.Membership = event.MembershipJoin
 		}
 		if member.PowerLevel != nil {
 			powerChanged = currentPower.EnsureUserLevelAs(portal.Bridge.Bot.GetMXID(), extraUserID, *member.PowerLevel) || powerChanged
 		}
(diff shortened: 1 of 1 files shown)


bridgev2/portal.go
```diff
@@ -4735,10 +4739,17 @@ func (portal *Portal) syncParticipants(
 			Displayname: currentMember.Displayname,
 			AvatarURL:   currentMember.AvatarURL,
 		}
+		if ghost != nil && ok {
+			content.Displayname = ghost.Name
+			if ghost.AvatarMXC != "" {
+				content.AvatarURL = ghost.AvatarMXC
```
Fizzadar, line 4745: Should we also be checking avatar? This function is already pretty complicated.. 

Perhaps there's a nicer way to do this and separate membership sync from profile.

tulir:
Member events should never drift 🤔

Fizzadar:
> Member events should never drift 🤔

I think the problem is specific to gmessages - the ghost names fallback to phone number if not found and then later we get the name. Though now I type that I _think_ that should trigger a profile update anyway which the HS would then propagate to the rooms.

In which case maybe it's a nameSet=true comparison issue or something...

highesttt:
I also have a similar problem with the snapchat bridge and this seems very similar to my patch

Fizzadar:
Tulir and I spent some time poking at this yesterday and concluded the primary issue here is a bug in our hungryserv homeserver 🫠 

Confirmed (at least cannot find) there's no races during room bootstrap going on, and also uncovered a gmessages bug too: https://github.com/mautrix/gmessages/pull/80.

PR #541, opened 2026-08-04
bridgev2/portal: fix stale room capability state

Fizzadar:
Capabilities were computed before ExtraUpdates applied the connector's freshly synced chat metadata, so the published capability set was always one metadata change behind, and the 24h throttle then dropped the correction entirely rather than deferring it.

- move the `UpdateCapabilities` call in `UpdateInfo` to after `ExtraUpdates`
- remove the 24h throttle, as the capability ID comparison already prevents redundant state events, and getting them is cheap


tulir (approved):
Probably makes sense

PR #549, opened 2026-08-14
bridgev2/matrix: Return correct error from clienthttp

radon-at-beeper:
In the case that an error occurred while processing a clienthttp response, bridgev2 would always incorrectly return ErrLoginCancelled instead of the actual error, because the step runner would see that it was returning an error, and cancel the context, which the clienthttp code assumed was a user cancellation even though it wasn't.


Fizzadar (approved):
Great spot, LGTM!

PR #551, opened 2026-08-18
crypto: support MSC4385 for pushing secrets to other devices

SpiritCroc:
### Checklist

### DIFF
--- a/crypto/sharing.go
+++ b/crypto/sharing.go
@@ -199,10 +199,8 @@ func (mach *OlmMachine) receiveSecretPush(ctx context.Context, evt *DecryptedOlm
 		Stringer("secret_name", content.Name).
 		Logger()
 
-	log.Trace().Msg("Handling secret push")
-
 	if evt.Sender != mach.Client.UserID {
-		log.Warn().Msg("Secret push was not from our own user, ignoring")
+		log.Debug().Msg("Ignoring secret push from another user")
 		return
 	} else if content.Name == "" || content.Secret == "" {
 		log.Warn().Msg("Ignoring secret push with empty name or value")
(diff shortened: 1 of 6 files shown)


crypto/sharing.go
```diff
@@ -188,3 +189,102 @@ func (mach *OlmMachine) receiveSecret(ctx context.Context, evt *DecryptedOlmEven
...
+		}
+		trust, err := mach.ResolveTrustContext(ctx, device)
+		if err != nil {
+			log.Err(err).Stringer("device_id", deviceID).Msg("Failed to resolve trust of device, not pushing secret to it")
+			continue
+		} else if trust < id.TrustStateCrossSignedVerified {
+			log.Debug().Stringer("device_id", deviceID).Stringer("trust_state", trust).Msg("Not pushing secret to unverified device")
+			continue
+		}
+		if err := mach.SendEncryptedToDevice(ctx, device, event.ToDeviceSecretPush, content); err != nil {
```
tulir: Sending the to-device events one by one seems fairly inefficient

crypto/sharing.go
```diff
@@ -188,3 +189,102 @@ func (mach *OlmMachine) receiveSecret(ctx context.Context, evt *DecryptedOlmEven
...
+		log.Debug().Msg("No secret push receiver configured, dropping pushed secret")
+		return
+	}
+	mach.SecretPushReceiver(ctx, evt, content)
+}
+
+// resolveTrustRefreshingKeys resolves device trust, re-fetching the user's device keys once if
+// the device doesn't resolve as cross-signed (its signature may be newer than our cached
+// device list, e.g. right after the device verified itself).
+func (mach *OlmMachine) resolveTrustRefreshingKeys(ctx context.Context, device *id.Device) (id.TrustState, error) {
```
tulir: Clients like sdk and gomuks generally don't want networking to happen in the sync loop, which makes this slightly sub-optimal. Might be better to leave resolving trust as something the callback has to do

crypto/sharing.go
tulir: `EncryptToDevices` + `Client.SendToDevice` might be the proper method to use here (after the existing trust state checking). Apparently I had added that for widget support in gomuks and it's not used anywhere else right now, but the code looks like it does the right thing for multi-device encryption

tulir (approved):
I didn't test the new bundled key signature checking code, but hopefully it works

PR #552, opened 2026-08-19
Safter portal deletes

Fizzadar:
To rebase merge


Fizzadar:
Not worth the extra logic vs. a delete queue. Cherry picked the second fix into main

PR #560, opened 2026-08-27
bridgev2/matrix: allow canceling a single step

radon-at-beeper:


tulir (approved):
Looks correct, we should probably write unit tests for the provisioning login state machine at some point

PR #563, opened 2026-08-31
appservice/intent: allow setting avatar when profile field is missing

PolynomialDivision:
Homeservers may return M_NOT_FOUND when avatar_url has not been set for a profile yet, for example Tuwunel. SetAvatarURL previously treated this valid response as a fatal error and returned before sending the new avatar.

Treat M_NOT_FOUND as an empty avatar and continue with the PUT request. All other errors remain fatal.

### Checklist

### DIFF
--- a/appservice/intent.go
+++ b/appservice/intent.go
@@ -502,9 +502,9 @@ func (intent *IntentAPI) SetAvatarURL(ctx context.Context, avatarURL id.ContentU
 		return err
 	}
 	resp, err := intent.Client.GetOwnAvatarURL(ctx)
-	if err != nil {
+	if err != nil && !errors.Is(err, mautrix.MNotFound) {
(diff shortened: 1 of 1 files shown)


PolynomialDivision:
@jevolk Thanks for adding the tuwunel fix. However, this PR should still be merged because your commit is just a workaround?

tulir:
The spec might get changed instead to define that displayname/avatar_url return 200 even when unset, because that was the behavior in all servers pre-MSC4133: https://matrix.to/#/!0KNSXYXB_2xtEUkQ9MGBRy5oNIOfAKoq2uIqPZCJbI8/%24DILQ8eUyvHbDB9eXFNQ-jnDYN0XcH9LM4yf_5D8Be1s?via=maunium.net&via=matrix.org&via=element.io

PolynomialDivision:
Okay, so I will close?

PR #564, opened 2026-09-04
[DESK-23336] bridgev2: add min/max length hints to login input fields

SilverBirchh:
Clients only receive a regex pattern for user input fields, to know how many characters to accept they would have to parse the regex. Bridges that ask for fixed-length codes can now say so directly, letting clients cap the input and render a proper code field.

### Checklist

### DIFF
--- a/bridgev2/login.go
+++ b/bridgev2/login.go
@@ -315,8 +315,8 @@ type LoginInputDataField struct {
 	// A regex pattern that the client can use to validate input client-side.
 	Pattern string `json:"pattern,omitempty"`
 	// Minimum and maximum input length as a hint for the client
-	MinLength int `json:"min_length,omitempty"`
-	MaxLength int `json:"max_length,omitempty"`
+	MinLength int `json:"min_length,omitzero"`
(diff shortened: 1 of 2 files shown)


bridgev2/login.go
```diff
@@ -314,6 +314,9 @@ type LoginInputDataField struct {
 	DefaultValue string `json:"default_value,omitempty"`
 	// A regex pattern that the client can use to validate input client-side.
 	Pattern string `json:"pattern,omitempty"`
+	// Minimum and maximum input length as a hint for the client
+	MinLength int `json:"min_length,omitempty"`
+	MaxLength int `json:"max_length,omitempty"`
```
tulir: omitempty only omits null and empty objects/arrays in jsonv2, so to prepare for that, ints and booleans need to use omitzero instead (which omits zero values defined by go, like 0, false, nil and probably empty non-pointer structs too)

They should also be added to the openapi doc (`./matrix/provisioning.yaml`)

```suggestion
	MinLength int `json:"min_length,omitzero"`
	MaxLength int `json:"max_length,omitzero"`
```

PR #565, opened 2026-09-08
bridgev2: add disappearing message mode that starts after the recipient reads it

iFixRobots:
Adds `after_read_by_recipient` for direct chats where a message's timer starts when its recipient reads it.

Outgoing messages wait for the other person's read receipt. Incoming messages wait for our own read receipt. The existing storage and scheduler handle expiry; `after_read` and `after_send` keep their current timer behavior.

Timer storage stays in the normal serialized event handler. There is no global capability, new completion hook, or separate transaction path.

Affected-package build, vet and existing tests pass locally. Earlier native checks confirmed recipient-read timing in both directions; this type-based revision has not yet been exercised end to end. Clients must recognize the new type before a connector exposes it.


tulir (approved):
Looks about right

PR #570, opened 2026-09-17
bridgev2: support recipient-read timers in groups

iFixRobots:
Extends `after_read_by_recipient` to group chats. Incoming messages wait for our own read receipt; outgoing messages wait for another member's read receipt.

Uses the receipt timestamp rather than the time it was processed, so delayed receipts don't extend the timer. Existing storage and scheduling are unchanged.

Verified incoming and outgoing group expiry locally. Build and vet passed.

Related: #565


iFixRobots:
@tulir - this is to finalize support for disappearing messages for both DMs and Groups in Viber. See the doc I sent you.

bridgev2/portal.go
```diff
@@ -1023,16 +1023,32 @@ func (portal *Portal) callReadReceiptHandler(
...
+	}
+	if portal.RoomType != database.RoomTypeDefault && portal.RoomType != database.RoomTypeGroupDM {
+		return
+	}
+	api, ok := source.Client.(NetworkAPIWithUserID)
+	if !ok {
+		return
+	}
+	userID := api.GetUserID()
+	if userID == "" || !sender.IsFromMe && (sender.Sender == "" || sender.Sender == userID || source.Client.IsThisUser(ctx, sender.Sender)) {
```
tulir: The second part of the check (everything except `userID == ""`) looks redundant? `IsFromMe` should always be accurate, so `Sender` should never be the login's own user id when IsFromMe is false

tulir (approved):

PR #574, opened 2026-09-21
bridgev2: preserve login state and message mappings

iFixRobots:
Keep credential and profile updates from overwriting a concurrently created filtering space. Preserve message mappings when redaction or persistence fails.

Add bounded media downloads and clear marked-unread state for own read receipts. Needed by the [WeChat SDK integration](https://github.com/beeper/beeper/pull/5000).


bridgev2/portal.go
```diff
@@ -1401,10 +1401,14 @@ func (portal *Portal) handleMatrixMessage(ctx context.Context, sender *UserLogin
...
 			portal.Bridge.GetGhostByID(ctx, message.SenderID)
-			err = portal.Bridge.DB.Message.Insert(ctx, message)
-			if err != nil {
-				log.Err(err).Msg("Failed to save message to database")
-			} else if resp.PostSave != nil {
+			if message.RowID == 0 {
+				err = portal.Bridge.DB.Message.Insert(ctx, message)
+				if err != nil {
+					log.Err(err).Msg("Failed to save message to database")
+					return EventHandlingResultFailed.WithMSSError(fmt.Errorf("%w: failed to save message mapping: %w", ErrDatabaseError, err))
```
tulir, line 1408: This is wrong: the message is already bridged at this point, so this line is always guaranteed to be a false negative

bridgev2/portal.go
```diff
@@ -1401,10 +1401,14 @@ func (portal *Portal) handleMatrixMessage(ctx context.Context, sender *UserLogin
 			// Hack to ensure the ghost row exists
 			// TODO move to better place (like login)
 			portal.Bridge.GetGhostByID(ctx, message.SenderID)
-			err = portal.Bridge.DB.Message.Insert(ctx, message)
-			if err != nil {
-				log.Err(err).Msg("Failed to save message to database")
-			} else if resp.PostSave != nil {
+			if message.RowID == 0 {
```
tulir, line 1404: Is the idea here to save the message in the network connector? That seems quite unusual and probably unnecessary, but maybe not entirely wrong

bridgev2/matrixinterface.go
```diff
@@ -225,6 +225,11 @@ type MatrixAPI interface {
 	GetEvent(ctx context.Context, roomID id.RoomID, eventID id.EventID) (*event.Event, error)
 }
 
+type MatrixAPIWithMediaLimit interface {
+	MatrixAPI
+	DownloadMediaWithLimit(ctx context.Context, uri id.ContentURIString, file *event.EncryptedFileInfo, maxBytes int64) ([]byte, error)
+}
```
tulir, line 231: This looks unnecessary, outbound file sizes are limited with room capabilities.

bridgev2/matrix/intent.go
```diff
@@ -153,14 +153,14 @@ func (as *ASIntent) MarkRead(ctx context.Context, roomID id.RoomID, eventID id.E
 		req.FullyRead = eventID
 		req.BeeperFullyReadExtra = extraData
 	}
-	if as.Matrix.IsCustomPuppet && as.Connector.SpecVersions.Supports(mautrix.BeeperFeatureInboxState) && as.Connector.Config.Homeserver.Software != bridgeconfig.SoftwareHungry {
+	if as.Matrix.IsCustomPuppet && as.Connector.SpecVersions.Supports(mautrix.BeeperFeatureInboxState) {
 		err = as.Matrix.SetBeeperInboxState(ctx, roomID, &mautrix.ReqSetBeeperInboxState{
-			//MarkedUnread: ptr.Ptr(false),
-			ReadMarkers: &req,
+			MarkedUnread: new(bool),
+			ReadMarkers:  &req,
```
tulir, line 159: IIRC using the inbox state endpoint causes bridges to mess with archive state, which we don't want, so probably can't do this. Mark unread bridging was probably also disabled for similar reasons

bridgev2/space.go
tulir, line 1: All the space and profile related changes look unnecessary, bridges shouldn't really care about any of that

bridgev2/portal.go
```diff
@@ -3776,11 +3780,14 @@ func (portal *Portal) handleRemoteMessageRemove(ctx context.Context, source *Use
 	dontRenderPlaceholderProvider, ok := evt.(RemoteMessageRemoveWithoutPlaceholder)
 	dontRenderPlaceholder := ok && dontRenderPlaceholderProvider.DontRenderPlaceholder()
 	res := portal.redactMessageParts(ctx, targetParts, intent, getEventTS(evt), "", dontRenderPlaceholder)
+	if !res.Success {
+		return res
+	}
 	err = portal.Bridge.DB.Message.DeleteAllParts(ctx, portal.Receiver, targetParts[0].ID)
```
tulir, line 3786: Hmm, this part is slightly more complicated than others. It might be fine, though it can also be incorrect in some cases. Probably doesn't make much difference either way

tulir (commented):
Most of the changes here are wrong/unnecessary

PR #575, opened 2026-09-21
bridgev2: retry saving newly created rooms

iFixRobots:
Let room creation retry its database save in the same process, reusing the room it already created. A failed save currently leaves MXID set in memory, so later attempts skip the save and remaining setup.


tulir:
The database shouldn't fail in any case except when the bridge is being stopped (and if that happens, the in-memory stuff is going to be dropped soon afterwards anyway)

PR #576, opened 2026-09-22
bridgev2: reset backfill completion when more history arrives

iFixRobots:
A completed backfill task can receive another manual batch with `HasMore=true`. The old completion flags survive that response, leaving the task marked done even though more history is available.

Set both completion flags from the response. Slow-fetch handling, terminal cutoffs and batch limits stay unchanged.

Build and vet passed locally for `./bridgev2/...` with `goolm`. All five CI jobs passed, including builds with both crypto backends and supported Go versions.


tulir (approved):
Most networks shouldn't need this, but it also shouldn't break anything on the networks that don't need it, so probably fine

